Issue #2, opened 2015-12-30
Walkback attempting to version all packages


jgfoster:
StsAdminShell>>#'versionSnapshotPackagesList:' references missing ProgressDialog 'Timer view'.

jgfoster:
Fix has been merged. As discussed, it is a bit crude but now we can version packages.

PR #7, opened 2016-01-01
Fix walkback in StsAdminShell when versioning all packages (issue #2).

jgfoster:
Also add ./Repository to .gitignore file.
Changes isolated in a branch.


jgfoster:
Hi Andy,

Thanks for your phenomenal gift to the Smalltalk community and for letting me contribute in a small way.

Knowing that fixing up STS is not a high priority makes sense and in a counter-intuitive way makes me feel better about making changes to get it working (I don't have to match the original author's style since my work can be just thrown away!).

I appreciate the feedback on the coding style (comments) and am happy to comply. I was initially cautious about the change but knowing it can be reviewed easily without comments makes sense. I believe I'm ready for another code review! (Also, any advice on version numbers for packages?)

As to AppVeyor, I believe Sebastian Calvo set it up (see https://groups.google.com/forum/#!topic/comp.lang.smalltalk.dolphin/Bz43AGRP7VI). The Git workflow is a bit new to me as well but I appreciate getting involved.

James

objectarts:
James,

I set up AppVeyor too, so I guess the build that was run with the Pull Request was mine. I just hadn't expected it.

PR #9, opened 2016-01-01
Allocate new GUIDs for Compiler COM interfaces to avoid conflict with…

objectarts:
… previous version of compiler.


jgfoster:
I see that you deleted and then restored the branch after the merge. I get the impression that deleting is a somewhat common practice to reduce clutter (and since it can be restored causes no loss of information). The advantage is reduced clutter. I'm curious as to your thought process for restoring.

objectarts:
I agree. I only restored it so I could find out how to delete it after that point. I just haven't done it yet. I'll do that now.

PR #14, opened 2016-01-05
Disable save system package

jgfoster:
Modifications to the system package ('Dolphin') cannot be saved and attempts to do so are silently ignored. This branch contains a changed to disable the menu items that are otherwise ineffective. It also contains minor layout changes to the dialog used to collect change information so as to avoid cutting off labels.


objectarts:
James, this change seems good but I'm now trying to think why need the #isOAD check in PackageSelector>>isSaveablePackage at al. In D6 this was really just to stop people accidentally saving the huge Dolphin package that they wouldn't then be able to load again anyway. Now, however we have the boot process so changes to Dolphin can get loaded on the next boot. I'll do some tests and see if I can remove the #isOAD check.

jgfoster:
I agree about the #isOAD check but I actually ran into trouble with the earlier check of comparing to the systemPackage ('Dolphin'). As far as I can tell, I can't save this package from within the image. It seems to me that if I can modify it from within Smalltalk, I should be able to save my changes and then rebuild the image.

PR #15, opened 2016-01-06
Update IDB Goodies from 6a to 6b

jgfoster:
With the exception of two files (IDB Method History.pax and MethodHistoryBrowser.cls), the IDB goodies appear to be based on version 6a from 2005. This pull request provides version 6b from 2006 (including the two aforementioned files to keep things consistent). Note that Ian's web site and email are no longer valid so I'm submitting what I have since it seems later and I've been using it consistently for almost a decade.


garduino:
+ 1 !

Saludos / Regards,
Germán Arduino
www.arduinosoftware.com


2016-01-06 0:04 GMT-03:00 James Foster <notifications@github.com>:

PR #16, opened 2016-01-06
Add Solutions Software components and enhancements

jgfoster:
http://www.solutionsoft.co.uk/widgets/editablelistview/index.html contains some enhancements that I've found useful. I'm not sure how much you want to fold into the base, but it seems valuable to me!

### DIFF
--- a/Core/Contributions/Solutions Software/ELVDemoPerson.cls
+++ b/Core/Contributions/Solutions Software/ELVDemoPerson.cls
@@ -1,0 +1,62 @@
+"Filed out from Dolphin Smalltalk X6.1"!
+
+Object subclass: #ELVDemoPerson
+	instanceVariableNames: 'name dateOfBirth gender isMarried address'
+	classVariableNames: ''
+	poolDictionaries: ''
+	classInstanceVariableNames: ''!
+ELVDemoPerson guid: (GUID fromString: '{C8BC1041-CE4F-4B05-B5B2-6C5A35472382}')!
+ELVDemoPerson comment: ''!
+!ELVDemoPerson categoriesForClass!Unclassified! !
+!ELVDemoPerson methodsFor!
+
+address
+	^address!
+
+address: anObject
+	address := anObject!
+
+dateOfBirth
+	^dateOfBirth!
+
+dateOfBirth: anObject
+	dateOfBirth := anObject!
+
+gender
+	^gender!
+
+gender: anObject
+	gender := anObject!
+
+isMarried
+	^isMarried!
+
+isMarried: anObject
+	isMarried := anObject!
+
+name
+	^name!
+
+name: anObject
+	name := anObject! !
+!ELVDemoPerson categoriesFor: #address!accessing!private! !
+!ELVDemoPerson categoriesFor: #address:!accessing!private! !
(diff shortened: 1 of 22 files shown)


garduino:
+1 !

Saludos / Regards,
Germán Arduino
www.arduinosoftware.com


2016-01-06 3:18 GMT-03:00 James Foster <notifications@github.com>:

objectarts:
Hi James. The intention is to keep DolphinCoreProduct as clear of "stuff" as possible. It's intended to be a minimal system (and, if anything, we should be removing packages from it wherever we can). Could you please add the SolutionSoft packages to DolphinBaseProduct instead so they make it into DCE and DPRO but not DCORE.

jgfoster:
Thanks for the guidance on where to put things. I've got another checkin with that change.

fxgallego:
Hi guys,

Do you have plans to add all the historical goodies somewhere in the
project / other repo?
I remember dolphinmap.net, and always is down, it would be nice to publish
all again in pax format.
Udo Scheneider was the mantainer.

Best Regards
 Sebastian Calvo

2016-01-06 15:51 GMT-03:00 James Foster <notifications@github.com>:

krodelin:
Just started the old Dolphin Map Image again - but DNS/Reverse Proxy seems to be broken. So you have to access it directly:
http://87.106.75.206:8080/WikiDoc/WikiPage

But please note that no Goodies were actually hosted on the server - only links.

jgfoster:
Sebastian, I plan to add everything that (1) I legally can, (2) I find useful, and (3) Andy will accept. ;-)

objectarts:
James, Sebastian, I think we need another repo (or set of repos) to contain the various goodies that exist. I don't think it is correct to have all these loaded into the image. I have however preloaded some of IDB's tools in the latest checkin because they are so useful. Please lobby me if there's something that should be in-image though.

objectarts:
Udo, thanks for putting up the Dolphin Map. I think we should create a Wiki here on GitHub and transfer the (still relevant) pages.

fxgallego:
1. Good idea on transfer useful content to Dolphin Wiki at Github
2. Contributed Repo, maybe one for contributor is ok. Where? As a repo in
dolphinsmalltalk?
3. I've uploaded https://github.com/fxgallego/Chris-Uppal-Goodies.git by now
4. Where to put / discuss new feature / enhancement proposals? Currently
there are some remote tools like https://slack.com/. I'm not an slack user,
only to mention. Maybe a feature request issue is enough but I don't want
pollution on the main issue tracker.

Best Regards
Sebastian

BTW: Sorry by my misuse of your language folks.

2016-01-06 19:06 GMT-03:00 Object Arts <notifications@github.com>:

jgfoster:
See https://github.com/dolphinsmalltalk/Dolphin/wiki for a start on the wiki.

bernhardkohlhaas:
I am also wondering, if we should migrate the documentation at http://www.object-arts.com/downloads/docs/ to the WIKI. @objectarts would that be okay with you?

objectarts:
@bernhardkohlhaas .. yes migrating the docs to the Wiki is a great idea .. they are somewhat moribund as they are now since I no longer have the tool that was used to create them.

We need to be careful that the documentation doesn't get fragmented (it did rather on the old Wiki) though.

PR #19, opened 2016-01-06
Provide boot/build information inside image

jgfoster:
Now that we can build our own image there is a certain uncertainty about when it was build and what is in it. This pull request attempts to address that situation. There is a new menu item on the SmalltalkSystemShell under Help, just above "About Dolphin Smalltalk" named "Boot Log" that shows information about the build process. If you are using Atlassian SourceTree then it includes git information (and I'm sure it can be improved to cover other git tools). I find it useful!


fxgallego:
Cool. At work we use a tool shell that gets its background red if the
commit that originated the image is not the same commit of your local
sources. That's because you could overwrite others code with your old image
code (if you download the image from the CI server). You should read the
sha of your current local branch and compare with the one stamped in the
image (by the CI session manager). I've build a Git command line client to
get this info but there is a Git library too.

Best regards

2016-01-06 18:55 GMT-03:00 James Foster <notifications@github.com>:

objectarts:
James, sorry, but I'm not happy with this as it stands. I like the idea in general but there are a couple of issues:

* It doesn't work if the images are built from a ZIP directory (i.e. the user is not using Git). In this case it also gives errors in the CMD console.
* I don't like the reliance on SourceTree. 

I think anything we do should work with any Git tooling and work for ZIP downloads too although I can't see how this can be achieved just yet.

jgfoster:
Andy, I agree that this isn't a really great solution. While it doesn't provide as much information if SourceTree isn't installed, it does still create a file and capture the boot/build date and time. I tried it without SourceTree being visible and it didn't give an error in the CMD console, so I'd like to better understand what you are seeing. What Git tools are you using? My idea was to make it work for you and me (!) and not be any worse for others (they still get the build date/time), and then gradually refine it (or let others contribute their solutions).

fxgallego:
This is the screenshot from my work development image. In the window
caption we show info about the build.
Sincerely, we never look that info but the image stamp is used in the image
vs. sources validation tool that is very useful.


2016-01-06 20:33 GMT-03:00 James Foster <notifications@github.com>:

objectarts:
@jgfoster .. try a download as ZIP and then doing a BootAll.
@fxgallego .. I can't ses a screenshot in your message?

fxgallego:
@objectarts http://imgur.com/hEQjNdq

jgfoster:
Andy, I've updated my branch to verify that there is a git checkout before attempting to use git information and avoid an error when building from a zip download. The script is also extensible to allow for alternate paths to the git executable as other tools are used.

objectarts:
Thanks James.. if you join the Gitter chat you'll see that I've had an idea to grab this Git info from within the image while it's booting using a shell ouot from Dolphin to Git running as an external process. This will give more control I think. Will let you know if it works.

What is the exact version info you think is needed to be displayed by Dolphin.. the log you produce seems a bit verbose to me,

objectarts:
I've checked in an implementation that perform the git command in the image rather than in the CMD.

fxgallego:
Cool, tested works ok.
Thanks

2016-01-11 19:43 GMT-03:00 Object Arts <notifications@github.com>:

PR #20, opened 2016-01-07
Support creation of another product

jgfoster:
There are three subclasses of DolphinProduct included as part of the boot image and any of these three can be built with a command-line argument. To build another product using the same process requires loading a new subclass of DolphinProduct before the build can proceed. This pull request provides one such way of doing so. If the named product (argument 3) isn't found then argument 4 is treated as the path to a package to be loaded and then the search for the product is repeated. I've been able to build a Jade development image using this approach and it seems pretty simply. I'm not sure, however, if it is the recommended approach. Like all my contributions, I'd welcome feedback and guidance.

### DIFF
--- a/Boot.st
+++ b/Boot.st
@@ -53,7 +53,10 @@ productName := SessionManager current argv at: 3 ifAbsent: [ | p |
 	isPrompted := true.
 	p name].
 
-[product := Smalltalk at: productName asSymbol.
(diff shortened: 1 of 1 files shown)


objectarts:
The DolphinProduct approach was really for releasing the old D6 products from a fully populated image by paring them down rather than by booting them up (although both work). I think we need something better going forwards but your approach seems fine for now.

PR #28, opened 2016-01-11
Boot in subdirectory

jgfoster:
These changes allow a **userdev** to boot a development image in a subdirectory of Dolphin's root directory. Now all application development activity can be isolated from **coredev** activity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.chg
 *.ERRORS
+*.img
 *.img7
 *.pac
 *.sml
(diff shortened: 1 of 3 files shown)


jgfoster:
I'm investigating an alternative of updating the package paths on startup if the image has changed.

objectarts:
Be careful you don't build in anything that rules out Workflows A & B. That is, don't assume that the core sources always reside in the same folder as Dolphin7.exe. I think this needs to be optional. The idea discussed on Gitter where we have an option to create a new app directory might have legs. e.g. something like:

> dolphin7 DPRO -init \MyHome\MyApp -sharecore

Might copy DPRO.img7 to \MyHome\MyApp\MyApp.img7 and somehow configure it to refer to the shared core packages in the original location. BTW, I don't necessarily like these option names.. just food for thought.

jgfoster:

PR #29, opened 2016-01-11
Add demo workflow

jgfoster:
This is a follow-up to the earlier [pull request](https://github.com/dolphinsmalltalk/Dolphin/pull/28) and provides the components needed for a sample workflow. I've created it separately since its inclusion may be more controversial. That is, while I've found it to be a useful approach, I don't assume I've convinced others!


jgfoster:

PR #31, opened 2016-01-11
Build git version into image on boot.

objectarts:


jgfoster:
I don't see any git info in my development image. When I attempt to access git using DevelopmentSessionManager, I get an error and it appears that there is an invalid STARTUPINFO {STARTUPINFO does not understand #dwSize}. 
Also, why is git info limited to the development image? It seems that it would be useful in other images as well.

objectarts:
Can you post a walkback for:

> DevelopmentSessionManager git: 'describe'

when the current directory is set to the boot directory. Do you have git on your path?

jgfoster:
Ah, no, I do not have `git` on the path. I had avoided that requirement in my [cmd file](https://github.com/jgfoster/Dolphin/commit/e30237f327037d2d8dedb117ebdab536b8a49015).

PR #33, opened 2016-01-12
Maintain accurate references to packages when image is moved

jgfoster:
If the paths to the packages were valid before an image was moved, modify the paths so that they are still valid. (Offered for consideration of the workflow options.)

### DIFF
--- a/Core/Object Arts/Dolphin/Base/SessionManager.cls
+++ b/Core/Object Arts/Dolphin/Base/SessionManager.cls
@@ -338,6 +338,9 @@ imageFileName
 
 	^File default: self imagePath extension: self imageExtension!
 
+imageMovedFrom: oldBase to: newBase
+	"Private - Available for subclasses to override"!
+
 imagePath
 	"Answer the base file path on which the image, source and changes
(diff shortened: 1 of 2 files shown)


objectarts:
Does this cater for the situation I describe in my comment to #28? That is, does it allow for Worflows A & B where the core sources are not assumed to be shared.? Only Workflow C assumes  shared core sources. From what I can see, this change will fix up any image that is copied to a new directory to point back to the original sources is that not correct?

jgfoster:
Yes, you are right. In **Workflow A** I read that moving the image 'effectively (as it stands now) makes Dolphin think that all the core packages actually' are still under the image when they are not. I was thinking of *breaking the link to the packages* as a bug not a feature(!). In **Workflow B** I read that there was some new mechanism that prevented writing to the Core directory. Whatever that new mechanism was, I assume it could be applied in this case as well. (I did try to avoid a walkback if the Dolphin sources did not exist so if the connection was already broken the move would still work.)

Perhaps it's back to the drawing board...

jgfoster:

Issue #34, opened 2016-01-12
Need command line options parser in SessionManager


objectarts:
In order to be able to support command line tool integration with Dolphin (such as automated testing) we need to be able to parse the command line passed to Dolphin7.exe.

jgfoster:
I'm going to take a look at this.

objectarts:
Now complete with new CommandLine class.

PR #35, opened 2016-01-13
Parse command line options

jgfoster:
This is a proposal to address Issue #34. While any feedback is welcome, I'm primarily interested in what API is needed. Also, if it gets to the point of being considered for inclusion, then I'd welcome guidance on the packaging (I assume that it doesn't warrant a separate package and tests should go somewhere else). This does not yet support long options nor multiple options with the same name. In any case, the API is what is interesting and since all the tests pass, I'm throwing it open for discussion.

Note that I haven't modified the boot script to include this so it needs to be loaded manually.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/Command Line.pax
+++ b/Core/Object Arts/Dolphin/Base/Command Line.pax
@@ -1,0 +1,44 @@
+| package |
+package := Package name: 'Command Line'.
+package paxVersion: 1;
+	basicComment: ''.
+
+
+package classNames
+	add: #CommandLine;
+	add: #CommandLineTestCase;
+	yourself.
+
+package binaryGlobalNames: (Set new
+	yourself).
+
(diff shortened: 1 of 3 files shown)


objectarts:
James, this is just the ticket, thanks.
 
I shall work out how to load this into the core image and place the tests somewhere so they can be run automatically at build time. 

FYI, Dolphin would normally call the test case class CommandLineTest (rather than CommandLineTestCases). If you do this than you can use Ctrl+Alt+T in the browser to switch between the test case and the class under test. Also you can use Ctrl+Alt+R while looking at the class under test and the tests will be run in the background - you can see the traffic light results in the Unit Tests tab header without having to switch tab. However, I think that "Test", "Tests" and "TestCases" are common enough suffixes that all should be supported by Ctrl+Alt+R rather than just the former. I'll make it so in a commit just now.

PR #36, opened 2016-01-18
Add CommandLineTest to Dolphin Base Tests package

jgfoster:
So that these tests are run with automated build.


jgfoster:
Interestingly, the automated tests show one failure and one error. This might be due to the fact that I've included tests for enhancements that haven't been merged into the master branch. I wonder if the automated test script should quit with a non-zero exit code if the SUnit tests have errors or failures.

PR #39, opened 2016-01-18
Edits to DBOOT

jgfoster:
This pull request provides a way to edit the DBOOT image so that the remaining boot process is a bit simpler. It also opens a path to not just put more into DBOOT but also take things out. Commit e4d7726 shows the state before the change to DBOOT and commit de9d362 shows the state after (they were separated to facilitate inspection of the actual modification process).


objectarts:
@jgfoster - I just haven't had a change to review this yet.

objectarts:
@jgfoster - I still haven't tried this,  but the way I read the code, you are permanently modifying DBOOT.img in stages as changes are made over time (using BootBoot). In  principle this is a good idea but in practice I don't want to lose the ability to boot from the **original** DBOOT.img. If anything happens that forces us to rebuild DBOOT from scratch (heaven forbid as it would be difficult) then, if we don't have a record of ongoing changes, we'll be "stuffed" (as the saying goes).

I think if you want to do this "cleaning stage", then BootBoot.st needs to be kept as a record of all the stages that have gone on to get to this point. At some point in the future we may be able to dispense with the original DBOOT but at present I wouldn't be happy to do so,

jgfoster:
@objectarts, Thanks for taking a look. While there is a benefit to community contributions I recognize the demands it places on you to look at all the submissions! That is, your workload isn't necessarily decreased since you need to provide guidance to us as well.

I recognize the risk of painting oneself into a corner with changes to DBOOT and the need to have a clear trail. My thought was that `git` and GitHub would provide that record of the changes since you can simply return to a previous commit. I was careful to commit the pre-run script as well as the post-run script for just this reason. I think with that step being a requirement for these types of changes then the process ought to be safe, reliable, and reversible. I do think that this sort of cleanup will pay dividends going forward, but I respect your caution.

jgfoster:
I'm going to withdraw this for now since I've discovered problems. Basically, in order to save the DBOOT image loaded additional packages. This goes against the goal of reducing the boot image and means that additional packages, such as Dolphin Base, are now more difficult to change. This problem was exposed by the failing tests for CommandLine parsing. I still like the idea but it needs work!

PR #40, opened 2016-01-18
Add some more file types to be treated as crlf

jgfoster:
To avoid false-positive changes when package is saved.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,9 @@
 # corruption we must avoid these indices changing by always forcing use of CRLF line endings.
 *.sml eol=crlf 
 *.chg eol=crlf
+
+# If files are downloaded using git on Mac or Linux, then saving a package from Windows
+# will create a number of false-positive changes (files that differ only in line endings).
+*.pac eol=crlf
+*.pax eol=crlf
+*.st  eol=crlf


objectarts:
@jgfoster, I'm going to merge this in but I'm not totally happy with it. The existing crlf settings isolate files that look like text but actually aren't (because they are indexed into). What you are doing here is marking some text files to be treated differently under alternate operative systems. Intuitively this doesn't sound right to me .. really developers shouldn't be modifying files on non-Windows systems IMO. Anyway, we can see how this goes and always revert later if it causes problems.

jgfoster:
@objectarts, Thanks for trying it out. My machines are Macs and I do my Dolphin work in a VM (using Fusion). The problem is that for some reason SourceTree is much slower in Windows than on the Mac so I do my source code management on the Mac but my (Dolphin) development in Windows. I don't modify files on the Mac, but still end up with line-ending conflicts. In any case, this change means that we are marking these text files to be treated *the same* under alternate operating systems (not differently). No matter which operating system is used for `git`, the files should follow the Windows convention. But, of course, if it causes problems for others then reverting it is appropriate.

fxgallego:
I gave up with SourceTree and its performance problems.
Now enjoy using Git Extensions
http://sourceforge.net/projects/gitextensions/

Saludos


2016-01-19 16:30 GMT-03:00 James Foster <notifications@github.com>:

objectarts:
@jgfoster - aha - that's exactly the workflow I think should be discouraged :) I too work on a Mac under Parallels and, yes, I guess it is a bit slow. One thing worries me though, will we have to label all text files in this way (.txt etc)?

fxgallego:
@objectarts if you clone in mac you are in problems.
Maybe a good solution for the strange case of using a mac for Dolphin development could be configuring git.
git config --global core.autocrlf false

fxgallego:
I forget that the repo was commited with auto=true...

jgfoster:
@objectarts, My inclination would have been to simply set text=crlf (vs. auto) instead of doing them piecemeal, but that risks being overinclusive (and I didn't want to go too far). Given that .gitattributes is a hidden file it seems like gradual movement in a particular direction (adding desired lines) is not too expensive. My experience was making a change to one package (Dolphin Development) and having 200 files show as changed but actually have no changes.

objectarts:
I find all of this very confusing. @fxgallego  are you saying that @jgfoster 's change will cause problems?

fxgallego:
Yes, it's confusing. I'm on a windows computer and my files are all ok. For me the @jgfoster changes are inoffensive. I don´t know how paralells works but if you use a program that runs on mac to clone the Dolphin repo or you've cloned the repo with SourceTree for windows and after you starts to use SourceTree for mac then I don´t know what could happen. So if you think that maybe some other would work this way, maybe it's a good idea to annotate all the files.
From https://help.github.com/articles/dealing-with-line-endings/
text eol=crlf
Git will always convert line endings to CRLF on checkout. You should use this for files that must keep CRLF endings, even on OSX or Linux. For example, here is a Windows project that enforces CRLF line endings

objectarts:
Let's leave it for now but I still think the original solution (just listing SML and CHG files as crlf) was cleaner because actually those files were "special" and so deserved to be in a special case. PAX etc are just text files and I think GitHub should be allowed to work as it normally does with these. Using Mac and Windows in tandem seems a degenerate workflow and should be discouraged.

Just as a matter of interest @jgfoster can you include these changes in your global .gitattributes file instead?

objectarts:
Hmm... you may have noticed (by all the commits) that I'm struggling to get various Dolphin tests to run cleanly under AppVeyor. Initially, I was just removing the failing tests from the active test suite. However, just looking at them now I see that they are all to do with comparing literal strings that span multiple lines. I wonder if the CRLF issue is biting here as well?

fxgallego:
@objectarts probably. Which files?. Do you use parallels to run Dolphin and SourceTree for mac?.
Maybe it's only AppVeyor or you because the eol conversion happens in the client.

The SML and CHG are annotated in gitattributes to avoid problems when the project is downloaded (not cloned) from GitHub and when a git config is forced to work autocrlf=false (this is not a default setting).

jgfoster:
@objectarts, According to http://git-scm.com/docs/gitattributes, "Finally global and system-wide files are considered (they have the lowest precedence)." And, yes, I suspect that test failures that involve literal strings that span multiple lines indicate a CRLF issue. Can you briefly describe the difference between master and automated-build-tests? (Though I guess I could do a diff!) Is there something we can do to help investigate these issues? Other than the CommandLineTests I don't think I've committed any changes to tests so I doubt I've impacted the CRLF issues (though anything is possible!).

jgfoster:
And I wish I had included *.cls in the CRLF list!

PR #45, opened 2016-01-20
Fix problem if no image name provided

jgfoster:
#allButFirst gives walkback


FunctionPoint:
Hi James, that was quick! 
I can indeed confirm that it works now.
Thanks!
Richard

PR #47, opened 2016-01-21
Support for Wine

jgfoster:
Update winefix.st to work with Dolphin 7 and provide a script to build a directory for use in Wine.


objectarts:
@jgfoster, I'm just putting together a binary deployment of the standard development environment in AppVeyor. It seems to me that I can add a similar thing for the deployment of the Wine directory like you have above (all the copying of EXEs, DLLs, images etc). Let me take a look.

jgfoster:
@objectarts, That's fine. Mostly I just wanted to show that it was possible and clean up the file-in. I'm happy to have a different approach.

jgfoster:
We'll see what other approach is taken...

PR #48, opened 2016-01-22
Panic can be initiated without prompt

jgfoster:
So it can be called from a script like winefix.st.


objectarts:
James, this is fine (indeed I added such a thing a few changes back but then removed it because I no longer needed it). But could you change the name of #panicWithoutPrompt to something like #reset please. "Panic" is a response to an abnormal state the user gets in, whereas "reset" sound a bit more like it is an intentional part of doing something.

jgfoster:
I appreciate the discussion of method names. This is an important part of building an application or tools.
How is the name 'hardResetGUI'?

PR #49, opened 2016-01-22
Edits to DBOOT

jgfoster:
This pull request has two commits (initially). First shows the script that modifies DBOOT and the second has the modifications. In this effort we have been careful not to add any packages to DBOOT. Something pretty neat about this is that `git` shows the edits to DBOOT.sml so you can see that the changes.


objectarts:
@jgfoster, I was going to comment on your withdrawal of the previous PR but you've beaten me to it. I would like to see it work something like this (which I think is a bit similar to what you have here but not quite).

1. When a change to `DBOOT.img7` is required a **coredev** makes a change to the file `UpdatedBoot.st` (could be `BootBoot.st` if you like but the word "boot" always means starting from scratch to me .. which this is not).
1. The **coredev** then runs the command file `UpdateBoot.CMD` that takes the existing checked out DBOOT image and updates it to include the changes. 
>NB: I don't like the moving and renaming of ST files within any language that doesn't have a proper #ensure: feature so I'd like to avoid this in CMD if we can. Doing something and then "putting it back like it was" always seems like bad style to me I'm afraid.

1. The end result that can be committed is a new `DBOOT.img7` that includes the recent update that is present in `UpdatedBoot.st`. The history of `UpdatedBoot.st' kept in GitHub is enough to show all the stages of boot updating that have taken place at any stage. This will allow us to replay them should it become necessary.

1. When subsequent changes to DBOOT are required the **coredev** removes the existing contents of `UpdatedBoot.st' and replaces then with the new change. There should be a comment at the top of `UpdatedBoot.st` (that remains for each cycle) that describes this whole process.

A modification to this process, which might leave things harder to read but easier to recover from, is to leave the previous changes in `UpdatedBoot.st` but just comment them out. New changes can be added just to the bottom. Then the entire history is available at a glance. What do you think?

objectarts:
@jgfoster , do we know why these tests are failing? Can you run a `BootPRO.cmd` and `TestDPRO.cmd` locally to see if they pass.

jgfoster:
@objectarts, yes, I'll take a look. Thanks for the suggestions.

jgfoster:
Thanks again for the suggestions. I believe that the current submission is an improvement and welcome further comments.

jgfoster:
I'm looking at the failed tests...

jgfoster:
I'm going to withdraw this for now. I think I've got a lead on what is failing but to test it I'm likely to try some commits and I don't want them to show up over here. I'll submit again when I'm more confident. Thanks for your patience!

PR #50, opened 2016-01-22
Allow code block to be part of CommandLineOption

jgfoster:
One of the characteristics of the getopt() approach is that the options are processed in order and a given option can appear multiple times. With the changes included here we now have that capability in Dolphin. Instead of testing for various keys in a dictionary (where the original order is lost and each key can have only one value), we now have the ability to provide a code block with each option definition and the code block is evaluated in the order each option is encountered. This also allows us to avoid writing if statements--something that seems to me to indicate a good design. To see the impact, see the simplification in DevelopmentSessionManager>>#processCommandLine.

One can now file in chunks from multiple files and the file-in can be done both before and after creating a new image (with the -i option). This avoids the problems "fixed" earlier by changing the order of the option testing (I wanted the file-in to happen after the new image was created, but someone might want it the other way).

Also this allows creation of an image without providing a name, and 'Dolphin' is the default (since it is the default if Dolphin7.exe is launched without any arguments). To see one use try the following to create two images:

C:\Dolphin>Dolphin7 DPRO.img7 -q -i -i JadeDev -x

With this I believe I've addressed the major goals of a command line parsing tool. Refinements and fixes can certainly come, but I think it is basically feature complete (unless someone wants the getopt_long_only capability to have long options with a single - prefix). 

Comments and suggestions are welcome.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/CommandLine.cls
+++ b/Core/Object Arts/Dolphin/Base/CommandLine.cls
@@ -56,14 +56,29 @@ addOptionAllowingArgumentNamed: aString
 	parsingRules add: (CommandLineOption allowingArgumentNamed: aString).
 !
 
+addOptionAllowingArgumentNamed: aString  whenPresentDo: aBlock
+
+	parsingRules add: (CommandLineOption allowingArgumentNamed: aString block: aBlock).
+!
+
 addOptionNamed: aString
 
 	parsingRules add: (CommandLineOption named: aString).
 !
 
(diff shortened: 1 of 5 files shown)


objectarts:
James, great job. 

I'll merge this as is, but I have a couple of comments.

* I find the method selectors like `#addOptionAllowingArgumentNamed:whenPresentDo:` rather long. Could you remove the `Named` portion (I think it is redundant).

* Also `#valueBlocks` doesn't seem very intention revealing. Could we choose something else, `#process` or `#processOptions` for example?

PR #53, opened 2016-01-24
Support edits to DBOOT

jgfoster:
`UpdateBoot.cmd` uses the normal `Boot.st` but with `UpdateBoot` as the argument instead of the name of a DolphinProduct subclass. Boot.st has a special check for this and if found invokes the script `Core\Object Arts\Dolphin\Base\Boot\Boot.st`. This script uses a package version to keep track of the most recently completed update and looks for a subsequent update (thus, it can be rerun safely and we don't need to comment-out completed updates). These updates are named `Core/Object Arts/Dolphin/Base/Boot/BootNNN.st` where NNN is the update number.`Boot001.st` provides some of the basic cleanup that was removed from the original `Boot.st` and `Boot002.st` loads some of the modified classes. Having these updates in two files demonstrates the stepwise update process and the second gives an example of what is likely to be a more typical update. Subsequent updates are done by creating a new `BootNNN.st` file, running `UpdateBoot.cmd`, and committing the changes to `git`. 

The original `DBOOT.img7` and `DBOOT.sml` are included in `Core\Object Arts\Dolphin\Base\Boot\` so one can repeat the process. Note that subsequent runs might not behave *exactly* the same since referenced files may have changed. For example, `Boot002.st` includes the code
  `File fileIn.`
If, for example, `Boot003.st`had subsequent changes to `File.cls` then the `Boot002.st` run would load the File version associated with `Boot003.st`. This will almost always be safe, but one could construct situations where update 3 relied on a method added in update 2 but removed in update 4. Given that `git` contains all previous versions and one can checkout any point along the way, we can still recreate intermediate images in the unlikely event that this is needed.

I think that this addresses the concerns raised earlier and I welcome further comments and suggestions.


objectarts:
@jgfoster. In Boot.st is the delay required do you think?

    SourceManager default compressSources!
    (Delay forSeconds: 5) wait!
    SessionManager current quit!

jgfoster:
@objectarts, No, the delay is not required. I put it in because the shell closes quickly and I don't get to read the output showing which updates were applied. I'm happy to take the delay out if you like!

objectarts:
@jgfoster, no that's fine, if there's a reason for it, it can stay.

PR #54, opened 2016-01-31
Prepare image for upgrade to VS2015

blairmcg:
Remove some dependencies in the image on pre VS2015 CRT that won't be supported on the later CRTs. This still runs on VS2013.


blairmcg:
@jgfoster, Hmmm, for changes to the Dolphin package, wouldn't it make sense just to reload this on boot (2nd stage as now)? It seems error prone to have to create a patch for base class changes. I can see value in having an early stage patching mechanism that is easy to hook a small patch into, but it doesn't seem the best way to handle base class changes that can be loaded normally.
BTW the boot image is pretty minimal already, as we trimmed it down to essentials over the years. I'm not sure there is much you could take out that would leave you with a functioning system that is able to start up and load other packages.

jgfoster:
@blairmcg, I'm not following very well. I understood that it is impossible to reload the Dolphin package. We can reload individual classes, and this practice was introduced into Boot.st (by 08d2b8217bf8ee7c9df4443b76a6902770d071c7). I thought that Boot.st was a bit more complicated than necessary (especially if we keep adding to it) and asked about modifying DBOOT. On 13-Jan I was told (in gitter) that "The DBOOT image can't be updated (at least not without months of work)." Faced with that as a challenge (!) I had the impression that finding a way to modify DBOOT might be useful and set about to do so.

Having accomplished that task, I am happy to have the experts determine whether this capability is actually useful or not. My thought was that having Boot.st continue to grow was not so attractive (I think it was improved by 5069d229b27d869343739c192c89af6d22b1ee95). I imagined that DBOOT might be updated occasionally, perhaps as a monthly or quarterly task to clean cruft out of Boot.st.

But again, I'm happy to have other approaches used. It just seemed useful and fun!

blairmcg:
@jgfoster I think what Andy meant was that rebuilding it from scratch (which was the way it used to be updated) would be very difficult. I think what you've done is useful anyway (as a way of patching things up before proceeding with more regular package load), but for picking up updates to the base classes we can probably work out how to reload the base package. Quite possibly just loading it at the start of boot.st before any of the others will work (although there may be some issues caused by re-initialization). The original process of creating one of the production images from the boot images did not include loading the base package, because the boot image was rebuilt with the latest content, so it would have been redundant to reload it again. I can't think of a fundamental reason why it couldn't be reloaded though.

jgfoster:
Thanks. I had a feeling I had misunderstood something!

PR #56, opened 2016-02-02
Simplify test script

jgfoster:
The existing test script has an exit if there are test failures/errors while it does not exit if all the tests pass. With this commit we have consistent behavior for both success and failure.


objectarts:
@jgfoster I have checked in a change to get the tests working again for the VS2015 build which has introduced a merge conflict with this PR.

PR #61, opened 2016-02-04
Test script cleanup

jgfoster:
Not ready to merge: This includes an explicit test failure to ensure that failures are caught.


jgfoster:
I think this is ready now.

Issue #68, opened 2016-02-08
Make Dolphin Wine aware


objectarts:

blairmcg:
Rather than messing up the base class library with workarounds for a specific runtime, couldn't polymorphism of the ExternalLibrary hierarchy be used to keep many of these workarounds out of the BCL? i.e. create a wine specific subclass of the relevant ExternalLibrary, and then implementing the workaround in an override of whatever library call Wine doesn't correctly or which is not implemented? 
ThemeLibrary provides an example of this - the class itself is abstract, and has a concrete implementation for Vista and later, and a mock implementation for earlier OSs that don't support some of the theming API. ThemeLibrary class>>new is then a factory method that returns an instance of the correct concrete subclass.

blairmcg:
Closing this as I believe its been done (or at least done for now, since its been open for a year)

PR #70, opened 2016-02-09
Provide check for isWine

jgfoster:
This was a small part of my earlier Wine stuff that may help with #68.


objectarts:
Thanks, I had been using the registry but this is probably neater.

PR #71, opened 2016-02-09
Apply edits from WineFix.st to image

jgfoster:
Using the WineFix.st script by itself in a development image seems to be okay but trying to deploy from that gives errors. With these changes we can now deploy successfully.


jgfoster:
This may provide further progress on #68.

objectarts:
@jgfoster, this is already being done as a part of my wine-fix change.

PR #89, opened 2016-02-23
#isNilOrEmpty

jgfoster:
This method is added to Collection, Object, and UndefinedObject by a contribution but its use requires inclusion of that third-party package in the image and the runtime packaging process. It would be easier if it were in the base.

### DIFF
--- a/Core/Contributions/ITC Gorisek/Dialect Abstraction Layer.pax
+++ b/Core/Contributions/ITC Gorisek/Dialect Abstraction Layer.pax
@@ -40,7 +40,6 @@ package methodNames
 	add: #ByteArray -> #itcAsInteger;
 	add: #Character -> #itcAsInteger;
 	add: #Collection -> #conform:;
-	add: #Collection -> #isNilOrEmpty;
 	add: #Collection -> #itcGetAbsSum;
 	add: #Collection -> #itcGetMax;
 	add: #Collection -> #itcGetMin;
@@ -96,7 +95,6 @@ package methodNames
 	add: #Object -> #isBlockClosure;
 	add: #Object -> #isDictionaryOrLookupTable;
 	add: #Object -> #isFileStream;
-	add: #Object -> #isNilOrEmpty;
 	add: #Object -> #isNilOrZero;
 	add: #Object -> #itcAsInteger;
 	add: #Object -> #itcAsNumberOrZero;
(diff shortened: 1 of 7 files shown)


objectarts:
I'll think about this.

There's a smell to this message that I really don't like (and yes, I know it it appears in other Smalltalks). The clue to the smell is, why would I ever do this:

4 isNilOrEmpty

I don't see this has any business being implemented on Object. What does it mean? And, actually, from a philosophical point of view I don't really understand why is it on UndefinedObject either? You are asking something with no sense of arity whether it has an arity of zero. It's a bit like saying:

4 isEmpty

and we'd never dream of doing that, would we! It's in Dialect Abstraction Layer because that is David's stuff and he uses it for STS. I'll probably come around but let me mull it over.

jgfoster:
I appreciate the discussion. I don't feel strongly about it but it was in some old code I inherited and when I loaded that package it conflicted with David's package. At that moment it seemed to me that if two different libraries add it then it would be easier to have it in the base. 

The use-case is for objects that don't initialize string instance variables and want to see if there is a non-empty string. Having phrased it that way I can certainly sympathize with the "smell" objection. Though, just to push back gently on your example, would you ever write the following? ;-)

4 isNil

Though maybe you are saying that implementing it in UndefinedObject and Collection is sufficient?

objectarts:
Actually, I see that Pharo doesn't implement this but `#isEmptyOrNil` At least it only does it on Object and Collection though so it's better than our implementation of `#isNilOrEmpty`.

I believe there is a use for a DAL (Dialect Abstraction Layer) to help code written for other environments to run on Dolphin, but I've always seen the existing DAL as David's and haven't paid much attention to its contents. I think the best thing is to leave `#isNilOrEmpty` where it is and instead add `isEmptyOrNil` to join it in the DAL (maybe with a suitable comment that only the feeble minded would make use of it :-) ).

I did come across the exact situation as you - conflicts with loose methods in the DAL when loading in third party code. I think we may need to extend the packaging system to accept such clashes with a warning rather than an error and leave the original method in place. I reckon we can assume that a same named method in the same class will have identical functionality. We wouldn't be able to accept class name clashes in the same way though.

Going back to `4 isNil`. I assume you're asking why we should implement `#isNil` on Object. Well the answer is because we are checking if _any_ object is in a slot rather than specifically a collection. And it's needed for true lazy initialisation. When I've seen `#isEmptyOrNil` used in the past it has, as you, say been a part of a premature optimization of lazy initialisation. The believe being that leaving a nominally collection holding instance variable as nil is more efficient than setting it to be an empty collection. But since:

    myCollection
        ^myCollection ifNil: [myCollection := #()]

always reuses a single instance of the same empty array there is no space saving by leaving it as nil.

jgfoster:
Thanks for the discussion. The "smell" is bad enough to give me second thoughts.

PR #98, opened 2016-02-27
Blairmc/osver

blairmcg:


blairmcg:
I've realised I need to push a new VM tag and update the FetchVM script before I can complete this PR as the builds are going to fail to boot.

Issue #99, opened 2016-02-28
TipOfTheDay is based on obsolete Windows help engine - update or remove


blairmcg:

blairmcg:
Tip of the day hasn't worked since the release of Windows Vista. Either we should kill it, or it needs to be replaced.

objectarts:
I think it should go.

PR #101, opened 2016-02-28
Remove obsolete TipOfTheDay class #99

blairmcg:
The TipOfTheDay uses the old windows help engine, and is no longer
supported on modern Windows OSs


blairmcg:
Oops

blairmcg:
@objectarts, how do you feel about excising the TipOfTheDay (deletes the class and all associated menu commands, etc)?

objectarts:
Sigh. You know that means editing every bloomin' tool view in the system. I'll have a go at some point soonish.

Issue #105, opened 2016-03-02
Extract to Component Refactoring cries wolf on the need to move pool variables


blairmcg:
The "Extract/Move to Component refactoring" may need to add pool dictionary references to the 
target class(es). However, it generates a warning whether or not the target class(es) already 
reference the required pools. It also generates this warning multiple times in some cases (which may
be a different issue).

PR #106, opened 2016-03-03
"Extract to Component" cries wolf on the need to move pool variables …

blairmcg:
…#105

Check to see if all target classes already reference the pool, and only
raise the warning if this is not the case.


Core/Contributions/Refactory/Refactoring Browser/Refactorings/ExpandReferencedPoolsRefactoring.cls
```diff
@@ -1,4 +1,4 @@
-"Filed out from Dolphin Smalltalk XP 2002 release 5.05"!
```
blairmcg: It's been a while since anyone has been in here :-)

PR #108, opened 2016-03-03
Fix walkback on (EditableListView show) 

jgfoster:
Provide a default rowHeight.


Core/Object Arts/Dolphin/MVP/ViewTest.cls
```diff
@@ -10,6 +10,37 @@ ViewTest comment: ''!
...
+	^list!
+
+badResourceIdentifiers
+
+	^#(
+		'DatePresenter.Month view'	"#109"
+		'TextPresenter.Masked edit'	"#110"
+	)!
+
+testAllResourceIdentifiers
```
blairmcg: This test will take many seconds to run, and I don't think it is going to detect many regression that matter.d It's worth keeping, but I don't think it should be run as part of the regression test suite.

Core/Object Arts/Dolphin/MVP/ViewTest.cls
```diff
@@ -10,6 +10,36 @@ ViewTest comment: ''!
...
+
+badResourceIdentifiers
+
+	^#(
+	)!
+
+testAllResourceIdentifiers
+	"Verify that all views actually can be opened. See #badResourceIdentifiers for failures"
+
+	true ifTrue: [^self].		"https://github.com/dolphinsmalltalk/Dolphin/pull/108"
```
blairmcg: It would be useful to be able to tag tests so that they are excluded from the regression test runs wouldn't it. At the moment our only mechanism to do this is to move the tests to another package as loose method (e.g. there are packages for long running tests, known failing tests). However these are black holes in which things get lost. A simpler approach would be just to use method categories. Even better an SUnit TestSuite should be constructible from a BrowserEnvironment. I'll raise an issue for this.

Issue #121, opened 2016-03-05
ColorRef>>asRGB converts in-place rather than answering a new object


blairmcg:
ColorRef>>asRGB sends #becomeA: to itself, a very nasty side effect.

blairmcg:
Fixed with bc66f8c

PR #122, opened 2016-03-05
ColorRef>>asRGB converts in-place rather than answering a new object #121 

blairmcg:
Also Color none and default should not translate directly to RGB colours without context - these only have meaning in relation to the environment in which the color ref is being used.


blairmcg:
I'm going to add/modify unit tests for this. Closing the PR until ready, when I'll reopen.

PR #145, opened 2016-03-11
Fix local failures in PackageTest and ProcessTest

jgfoster:
Something about my environment means that I consistently get three test failures. This fixes them by allowing more time for a process to terminate and delaying longer between file operations if the file system has one-second granularity on timestamps. It also improves the UI updating during the SUnit tests and should not slow down the tests on unaffected platforms.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/DolphinTest.cls
+++ b/Core/Object Arts/Dolphin/Base/DolphinTest.cls
@@ -77,14 +77,6 @@ runKnownBug: aString test: aSymbol
 	#todo.	"Make this controllable by an option or class var"
 	^false!
 
-setUp
-
-	super setUp.
-	(SessionManager inputState)
-		pumpMessages;
-		processDeferredActions.
-!
-
 shouldRaiseZeroDivide: aBlock
 	| raised |
 	Float reset.
@@ -105,7 +97,6 @@ shouldRaiseZeroDivide: aBlock
 !DolphinTest categoriesFor: #isOnline!public!unit tests! !
 !DolphinTest categoriesFor: #methodContextClass!constants!private! !
 !DolphinTest categoriesFor: #runKnownBug:test:!public!unit tests! !
-!DolphinTest categoriesFor: #setUp!public!Running! !
 !DolphinTest categoriesFor: #shouldRaiseZeroDivide:!Accessing!public! !
 
 !DolphinTest class methodsFor!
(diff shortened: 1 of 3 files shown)


Core/Object Arts/Dolphin/Base/DolphinTest.cls
```diff
@@ -77,6 +77,14 @@ runKnownBug: aString test: aSymbol
 	#todo.	"Make this controllable by an option or class var"
 	^false!
 
+setUp
+
```
blairmcg: This seems an inappropriate change to make in the base class of all the Dolphin tests, many of which are completely unrelated to UI activity.

Core/Object Arts/Dolphin/Base/DolphinTest.cls
```diff
@@ -97,6 +105,7 @@ shouldRaiseZeroDivide: aBlock
 !DolphinTest categoriesFor: #isOnline!public!unit tests! !
 !DolphinTest categoriesFor: #methodContextClass!constants!private! !
 !DolphinTest categoriesFor: #runKnownBug:test:!public!unit tests! !
+!DolphinTest categoriesFor: #setUp!public! !
```
blairmcg: Requires categorisation

Core/Object Arts/Dolphin/Base/PackageTest.cls
```diff
@@ -363,7 +363,25 @@ testPaxTimestampUpdate
 	self deny: b isRenegade!
 
 testTimestampInitialized
-	self assert: Package new timestamp asInteger = 0! !
+	self assert: Package new timestamp asInteger = 0!
```
blairmcg: Rather than do this, let's just delete these two tests. I don't think they are worth the cost of running them.

Core/Object Arts/Dolphin/Base/DolphinTest.cls
jgfoster: I'm happy to take it out, but I'd point out that there isn't anything UI-specific about allowing background tasks to complete between tests. In particular, a test that forks a process and then expect it to complete running within a number of milliseconds will behave differently if it is run as a suite and if it is run in isolation. Further, when running non-UI tests from the TestRunner, it is frustrating to get no feedback. Finally, if there are no pending tasks/messages, then the overhead of checking should be minimal.

blairmcg: Yes, but it still doesn't belong here. I agree it would be useful if the TestRunner UI updated, but that should be done by observing the TestSuite change events. The UnitTestPlugin does this, so take a look at UnitTestPlugin>>runSuite: to see how to modify the TestRunner to achieve the same effect :-)

Issue #144, opened 2016-03-10
Interrupt key needs to be changed from Ctrl+Break as not available on many keyboards


blairmcg:
Mac keyboards, and also those on many laptops, do not have a Break key.

blairmcg:
There is still a place in the image that is looking for VK_CANCEL. Leaving this open until that is fixed.

PR #146, opened 2016-03-12
Interrupt key needs to be changed from Ctrl+Break #144

blairmcg:
Leave a comment


blairmcg:
Please note the default break key is now Ctrl+F12, although this can be changed by inspecting "Smalltalk developmentSystem" and changing the interruptKey aspect using the hotkey editor (i.e. you just press the key combination you want).

PR #160, opened 2016-03-15
Add ability to create #storeString for most views

jgfoster:
One of the (very few) frustrations of Dolphin is editing a view and having the only difference be changes in a massive literal array. One of the very nice features of VSE's WindowBuilder was that views were created by code so you could actually see what was changed (and even make edits directly in the code if you wanted).

With this commit, View>>#'storeString' will now return a string that when compiled will generate a similar view (with a few exceptions). I anticipate that there will be some comments and suggestions, and I look forward to the discussion!

### DIFF
--- a/Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
+++ b/Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
@@ -795,6 +795,29 @@ state
 	sequence list addAll: controlState afterIndex: 1.
 	^sequence!
 
+storeOn: aStream indent: anInteger variables: aDictionary
+
+	| default name |
+	name := self storeNameIn: aDictionary.
+	aStream display: name; cr.
+	anInteger + 1 timesRepeat: [aStream tab].
+	[
+		default := (self class progId: self progId) create; show.
+		aStream
+			nextPutAll: ' progId: ';
+			print: self progId;
+			nextPut: $; ;
+			cr.
+		anInteger + 1 timesRepeat: [aStream tab].
+		aStream nextPutAll: 'create;'.
+		self storeAttributesDifferingFrom: default On: aStream variables: aDictionary indent: anInteger + 1.
+		aStream cr.
+		anInteger + 1 timesRepeat: [aStream tab].
+		aStream nextPutAll: 'yourself.'; cr.
+	] ensure: [
+		default topView close.
+	].!
+
 supportedInterfaces
 	"Answer the set of interface classes supported by the receiver.
 	We need only support the standard property notification sink interface since
@@ -903,6 +926,7 @@ usesAmbientColors
 !AXControlSite categoriesFor: #showHatching!ambient properties!public!testing! !
 !AXControlSite categoriesFor: #sink!accessing!public! !
 !AXControlSite categoriesFor: #state!binary filing!private! !
+!AXControlSite categoriesFor: #storeOn:indent:variables:!printing!public! !
 !AXControlSite categoriesFor: #supportedInterfaces!constants!public! !
 !AXControlSite categoriesFor: #triggeredEvents!events!public! !
 !AXControlSite categoriesFor: #usesAmbientColors!public!testing! !
--- a/Core/Object Arts/Dolphin/Base/ArrayedCollection.cls
+++ b/Core/Object Arts/Dolphin/Base/ArrayedCollection.cls
@@ -121,7 +121,36 @@ sortUsing: aSortAlgorithm
 	aSortAlgorithm 
 		sort: self
 		from: 1
-		to: self size! !
+		to: self size!
+
+storeOn: aStream indent: anInteger variables: aDictionary
+	"Append to the <puttableStream> argument, target, an expression which when 
+	evaluated will answer a collection similar to the receiver."
+
+	(self allSatisfy: [:each | each isImmutable]) ifTrue: [
+		aStream print: self.
+		^self.
+	].
+	aStream
+		nextPutAll: '((';
+		nextPutAll: self class name;
+		nextPutAll: ' new: ';
+		print: self size;
+		nextPut: $).
+	1 to: self size do: [:i | 
+		aStream cr.
+		anInteger timesRepeat: [aStream tab].
+		aStream 
+			nextPutAll: 'at: ';
+			print: i;
+			nextPutAll: ' put: '.
+		(self at: i) storeOn: aStream indent: anInteger + 1 variables: aDictionary.
+		aStream nextPut: $;.
+	].
+	aStream cr.
+	anInteger timesRepeat: [aStream tab].
+	aStream nextPutAll: 'yourself)'
+! !
 !ArrayedCollection categoriesFor: #add:!adding!public! !
 !ArrayedCollection categoriesFor: #at:!accessing!public! !
 !ArrayedCollection categoriesFor: #at:put:!accessing!public! !
@@ -135,6 +164,7 @@ sortUsing: aSortAlgorithm
 !ArrayedCollection categoriesFor: #resize:!mutating!public! !
 !ArrayedCollection categoriesFor: #size!accessing!public! !
 !ArrayedCollection categoriesFor: #sortUsing:!public!sorting! !
+!ArrayedCollection categoriesFor: #storeOn:indent:variables:!printing!public! !
 
 !ArrayedCollection class methodsFor!
 
--- a/Core/Object Arts/Dolphin/Base/BlockClosure.cls
+++ b/Core/Object Arts/Dolphin/Base/BlockClosure.cls
@@ -549,6 +549,33 @@ stackTempCount: anInteger
 	info := (info bitAnd: ##((16rFF << 15) bitInvert)) 
 				bitOr: ((anInteger bitAnd: 16rFF) bitShift: 15)!
 
+storeOn: aStream
+	"Append, to aStream, a <String> whose characters can recreate the receiver."
+
+	| m block class selector string |
+	m := self method.
+	(m notNil and: [m isExpression]) ifTrue: [	"We can print the source"
+		aStream 
+			nextPut: $(;
+			nextPutAll: m getSource;
+			nextPut: $).
+		^self].
+	m ifNil: [^super storeOn: aStream].
+	class := m methodClass.
+	selector := m selector.
+	(class isKindOf: Class) ifTrue: [
+		string := class name , ' new ' , selector.
+		block := class new perform: selector.
+	] ifFalse: [
+		string := class instanceClass name , ' ' , selector.
+		block := class instanceClass perform: selector.
+	].
+	block = self ifTrue: [
+		aStream nextPutAll: string.
+	] ifFalse: [
+		super storeOn: aStream.
+	]!
+
 tempCount
 	"Answer the total <integer> number of temps used by the receiver. 
 	This includes all arguments, stack temps, copied values, and environment 
@@ -778,6 +805,7 @@ whileTrue: iterationBlock
 !BlockClosure categoriesFor: #repeat!evaluating!public! !
 !BlockClosure categoriesFor: #stackTempCount!accessing!private! !
 !BlockClosure categoriesFor: #stackTempCount:!accessing!development!private! !
+!BlockClosure categoriesFor: #storeOn:!development!printing!public! !
 !BlockClosure categoriesFor: #tempCount!accessing!public! !
 !BlockClosure categoriesFor: #value!evaluating!public! !
 !BlockClosure categoriesFor: #value:!evaluating!public! !
--- a/Core/Object Arts/Dolphin/Base/Class.cls
+++ b/Core/Object Arts/Dolphin/Base/Class.cls
@@ -1,4 +1,4 @@
-"Filed out from Dolphin Smalltalk X6.1"!
+"Filed out from Dolphin Smalltalk 7"!
 
 ClassDescription subclass: #Class
 	instanceVariableNames: 'name classPool sharedPools comment classCategories _guid '
@@ -603,6 +603,13 @@ storeGUID
 	self sourceManager storeGUIDForClass: self
 !
 
+storeOn: aStream
+
+	aStream
+		nextPutAll: '##(';
+		nextPutAll: self name;
+		nextPut: $)!
+
 subclass: aClassSymbol instanceVariableNames: instVarString classVariableNames: classVarString poolDictionaries: poolDictString
 	^(self 
 		newClassBuilder: aClassSymbol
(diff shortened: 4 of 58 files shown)


blairmcg:
This is going to take a while to review. I'll have to do it in stages.

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
```diff
@@ -795,6 +795,29 @@ state
 	sequence list addAll: controlState afterIndex: 1.
 	^sequence!
 
+storeOn: aStream indent: anInteger variables: aDictionary
+
+	| default name |
+	name := self storeNameIn: aDictionary.
+	aStream display: name; cr.
+	anInteger + 1 timesRepeat: [aStream tab].
```
blairmcg: Is it worth bothering with attempting a poor-man's pretty print? You could cut some code and get a better quality result by just throwing this through the parser and formatter when you want to produce the final result.
Incidentally a general comment I have is that any formatting style is allowed, as long as it is the one produced by the RBFormatter as configured in the image :-) (to get this, Ctrl+Shift+S to reformat accept the methods, or Ctrl+W to reformat then without saving when working).

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
```diff
@@ -795,6 +795,29 @@ state
...
+	[
+		default := (self class progId: self progId) create; show.
+		aStream
+			nextPutAll: ' progId: ';
```
blairmcg: A general approach we follow when "printing" code to a stream, is to display the actual selector symbol. This means that subsequently this will be found when browsing for references, and would mean that the code would participate in any refactoring operation such as a rename that changed the selector. This is slightly more awkward than using strings, but is worth it for the reference browsing alone.

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
```diff
@@ -795,6 +795,29 @@ state
...
 	^sequence!
 
+storeOn: aStream indent: anInteger variables: aDictionary
+
+	| default name |
+	name := self storeNameIn: aDictionary.
+	aStream display: name; cr.
+	anInteger + 1 timesRepeat: [aStream tab].
+	[
+		default := (self class progId: self progId) create; show.
```
blairmcg: Creating a new instance to determine default values seems rather undesirable.

Core/Object Arts/Dolphin/Base/ArrayedCollection.cls
```diff
@@ -121,7 +121,36 @@ sortUsing: aSortAlgorithm
...
 		sort: self
 		from: 1
-		to: self size! !
+		to: self size!
+
+storeOn: aStream indent: anInteger variables: aDictionary
+	"Append to the <puttableStream> argument, target, an expression which when 
+	evaluated will answer a collection similar to the receiver."
+
+	(self allSatisfy: [:each | each isImmutable]) ifTrue: [
```
blairmcg: isImmutable is not limited to things with literal form. You'll need to use #isLiteral for this purpose. In fact this is already correctly defined on the relevant collections (Array and ByteArray)

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
```diff
@@ -903,6 +926,7 @@ usesAmbientColors
 !AXControlSite categoriesFor: #showHatching!ambient properties!public!testing! !
 !AXControlSite categoriesFor: #sink!accessing!public! !
 !AXControlSite categoriesFor: #state!binary filing!private! !
+!AXControlSite categoriesFor: #storeOn:indent:variables:!printing!public! !
```
blairmcg: Put all these in a 'storing' method category, not printing.

Core/Object Arts/Dolphin/Base/BlockClosure.cls
```diff
@@ -549,6 +549,33 @@ stackTempCount: anInteger
 	info := (info bitAnd: ##((16rFF << 15) bitInvert)) 
 				bitOr: ((anInteger bitAnd: 16rFF) bitShift: 15)!
 
+storeOn: aStream
+	"Append, to aStream, a <String> whose characters can recreate the receiver."
+
+	| m block class selector string |
+	m := self method.
+	(m notNil and: [m isExpression]) ifTrue: [	"We can print the source"
```
blairmcg: Source form could also be used for any completely clean block (any that are non-full and non-copying, and don't need self), as these can be taken out of the method context. I'm not sure how easy it is to find the source range, however, at least not without using the parser.

Core/Object Arts/Dolphin/Base/BlockClosure.cls
```diff
@@ -549,6 +549,33 @@ stackTempCount: anInteger
...
+		aStream 
+			nextPut: $(;
+			nextPutAll: m getSource;
+			nextPut: $).
+		^self].
+	m ifNil: [^super storeOn: aStream].
+	class := m methodClass.
+	selector := m selector.
+	(class isKindOf: Class) ifTrue: [
```
blairmcg: Instantiating objects and evaluating arbitrary methods could have unknown and undesirable side effects. This doesn't seem appropriate for a general purpose #storeOn: implementation, so I don't think you can do this. Also there are many cases where this could fail (object cannot be instantiated with new, not a zero arg selector).

Core/Object Arts/Dolphin/Base/Class.cls
```diff
@@ -603,6 +603,13 @@ storeGUID
 	self sourceManager storeGUIDForClass: self
 !
 
+storeOn: aStream
+
+	aStream
+		nextPutAll: '##(';
+		nextPutAll: self name;
```
blairmcg: display: self

Core/Object Arts/Dolphin/Base/Date.cls
```diff
@@ -277,12 +277,14 @@ storeOn: aStream
...
 
-	aStream
-		display: self class;
-		space;
-		display: #fromDays:;
-		space;
-		display: self asDays!
+	aStream 
+		nextPut: $(;
+		nextPutAll: self class name;
```
blairmcg: similarly, display: self class

Core/Object Arts/Dolphin/Base/Message.cls
```diff
@@ -107,6 +107,32 @@ setSelector: aSymbol arguments: argArray
 	selector := aSymbol.
 	args := argArray!
 
+storeOn: aStream
+
+	self argumentCount == 0 ifTrue: [
+		aStream
+			nextPutAll: '(Message selector: ';
```
blairmcg: As with the selectors, emit them directly from the object itself so that the reference is visible to tooling, rather than using a string.

Core/Object Arts/Dolphin/Base/Message.cls
```diff
@@ -107,6 +107,32 @@ setSelector: aSymbol arguments: argArray
...
+	].
+	aStream
+		nextPut: $(;
+		nextPutAll: self class name;
+		nextPutAll: ' selector: ';
+		print: selector.
+	0 < self argumentCount ifTrue: [
+		1 == self argumentCount ifTrue: [
+			aStream nextPutAll: ' argument: '.
+			(args at: 1) storeOn: aStream.
```
blairmcg: CodeMentor warning: args first

Core/Object Arts/Dolphin/Base/MessageSendAbstract.cls
```diff
@@ -99,6 +99,28 @@ receiver: anObject
...
+
+	aStream
+		nextPut: $(;
+		nextPutAll: self class name;
+		nextPutAll: ' receiver: '.
+	self receiver storeOn: aStream.
+	aStream 
+		nextPutAll: ' selector: ';
+		print: selector.
+	0 < self argumentCount ifTrue: [
```
blairmcg: General comment: We don't use this c-style inverted comparison. New code needs to be consistent with the existing conventions

Core/Object Arts/Dolphin/Base/Metaclass.cls
```diff
@@ -164,7 +164,15 @@ sharedStaticPools
...
 	"Save out a binary representation of the receiver to anSTBOutFiler."
 
-	anSTBOutFiler saveObject: self as: (STBMetaclassProxy forClass: self)! !
+	anSTBOutFiler saveObject: self as: (STBMetaclassProxy forClass: self)!
+
+storeOn: aStream
+
+	aStream
+		nextPutAll: '##(';
+		nextPutAll: self name;
```
blairmcg: print: self

Core/Object Arts/Dolphin/Base/Object.cls
```diff
@@ -1319,11 +1319,61 @@ stbSaveOn: anSTBOutFiler
...
+		at: #'commandDescription' 	put: [:anObject | aStream nextPutAll: 'command: '; print: anObject command];
+		at: #'buddy'			 	put: [:anObject | 
+			self isAutoBuddy ifTrue: [
+				aStream nextPutAll: 'isAutoBuddy: true'
+			] ifFalse: [
+				aStream nextPutAll: 'buddy: '; nextPutAll: (aDictionary at: anObject)
+			]
+		];
+		yourself.
+	aspects := self class publishedAspectsOfInstances asSortedCollection.
```
blairmcg: #publishedAspects is a development system package concept. You simply can't use this in the base. This method assumes too much knowledge of specifics of particular classes anyway to be appropriate as an Object method.

Core/Object Arts/Dolphin/Base/Object.cls
```diff
@@ -1319,11 +1319,61 @@ stbSaveOn: anSTBOutFiler
...
+	aspects := aspects reject: [:each | rejects includes: each name].
+	aspects do: [:each | | x |
+		(x := self perform: each name) = (default perform: each name) ifFalse: [
```
blairmcg: Aspects provide methods to read the aspect values

Core/Object Arts/Dolphin/Base/Object.cls
```diff
@@ -1319,11 +1319,61 @@ stbSaveOn: anSTBOutFiler
...
+	aStream nextPutAll: 'yourself)'
+!
+
 storeOn: aStream
 	"Append to the <puttableStream> argument, target, an expression which when 
 	evaluated will answer a collection similar to the receiver."
 
-	self printOn: aStream!
+	aStream nextPutAll: '(Object fromLiteralStoreArray: '.
+	self literalStoreArray storeOn: aStream.
```
blairmcg: literalStoreOn:

Core/Object Arts/Dolphin/MVP/Type Converters/NullConverter.cls
```diff
@@ -10,6 +10,16 @@ NullConverter comment: 'NullConverter is a <typeConverter> which applies no tran
 !NullConverter categoriesForClass!MVP-Type Converters-General! !
 !NullConverter methodsFor!
 
+= comparand
```
blairmcg: Seems the wrong level to introduce #hash and #=. These should be pushed up if they are needed.

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
jgfoster: Yes, I should use the built-in formatter.

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
jgfoster: Agreed.

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
jgfoster: The goal was to write code that looked as human-written as possible by leaving out the values that would otherwise be set by the initialize process. To the extent that it does create new instances, it is in the context of View, and the view composer does the same. On the other hand, if you'd prefer I can set all the values explicitly, even if they are the same as the default values; it is just cluttering the code with things that a human author would leave out.

Core/Object Arts/Dolphin/Base/ArrayedCollection.cls
jgfoster: Unfortunately, Array's implementation gets a stack overflow on cyclic references. I'll revisit this.

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
jgfoster: Should I change the categories of the existing #'storeString' and #'storeOn:' methods?

Core/Object Arts/Dolphin/Base/BlockClosure.cls
jgfoster: In existing views, the literal array creates a BlockClosure that names a source method but may be completely different so there isn't really any source anywhere.

jgfoster:
Thanks for taking the time to review this. How strongly do you prefer the JSON approach to the Smalltalk one? I'm not sure how well JSON handles cyclic references and I imagined that Dolphin users would find reading Smalltalk easier!

As to using storeString, should I try to salvage any of the more general work, or are you satisfied with how storeString works now (default is to answer a printString that typically can't recreate the receiver) and I should focus on isolating this to describing Views?

Core/Object Arts/Dolphin/Base/BlockClosure.cls
fxgallego: You are right. Sometimes there is no source so you need a decompiler or
keep the block as literal.

You should be aware that the VC allows users to touch trough the inspector
other values than those that can be edited via aspect editors.

Maybe you will encounter views larger than the maximum input string size
for the compiler.

Best Regards

jgfoster:
I'll leave storeString as-is and refactor this into a separate package that is aimed solely at View instances in a development image.

blairmcg:
Ok @jgfoster, thanks, I think that is the right choice.
I certainly don't think the current #storeOn:/storeString stuff is even vaguely useful or complete. This was only added at all because it was part of Blue Book Smalltalk, but only in a very half-hearted way. Some of your changes to that side of things are definitely improvements, but in the interests of trimming rather pointless fat from the base image I think it might be better if we removed it altogether rather than trying to fix it up. That would also make it easier for someone to provide a better implementation as an extension package, if it was needed for (say) to help run code from another dialect.
Nevertheless, I don't think it is a particularly good way to provide what is needed here.

Core/Object Arts/Dolphin/ActiveX/OCX/AXControlSite.cls
jgfoster: How far down the rabbit hole do I go with this? #new? #yourself? How do I handle keyword selectors with multiple arguments?

PR #162, opened 2016-03-17
Store string for view

jgfoster:
I'm still investigating a couple failing tests, but thought I'd submit this for review and comment.


blairmcg:
James, I think the objective behind this change (to provide a more readable representation for views that both diffs well and is potentially editable) is sound. But I don't think the approach is. The requirement is to be able to represent specific instances of views as configured by the view composer, which could include any kind of custom configuration of the view. The views were also not really designed nor implemented with side-effect free rehydration through code in mind, as this has always been done (thinks: I wonder if we have passed the 20 year anniversary yet?) through serialization rather than executing a description in code.  Constructing a code description of the view instances that is completely reliable and covers all the corner cases is possible, but would need to be backed up by a significant bulk of unit tests to be sure. It will also impose an ongoing maintenance effort. For these reasons I think the right solution to the original objective is to improve the literal filer representation to be more along the lines of json (i.e. a structured name/value pairs), and therefore I do not think this should be added into the booted dpro image, as I do not think we would want to change the standard view resource representation to use it. By all means put it under contrib if you wish.
@objectarts, what is your view?

Core/Object Arts/Dolphin/Installation Manager/DolphinProfessional.cls
```diff
@@ -78,6 +78,10 @@ contents
 		add: #('Core\Object Arts\Dolphin\MVP\Presenters\Prompters\Dolphin In-place Text Editor.pax' #encrypted #imageBased);
 		add: #('Core\Object Arts\Dolphin\IDE\Professional\Dolphin Refactoring Browser.pax' #encrypted #imageBased).
 
+	"Dolphin native Refactoring Browser integration"
```
blairmcg: As stated in my overall comment, I do not think this should be loaded by default, nor stored under IDE/Professional

blairmcg:
Thanks for moving it. I'll merge this one. 
I came across [STON ](https://ci.inria.fr/pharo-contribution/job/EnterprisePharoBook/lastSuccessfulBuild/artifact/book-result/STON/STON.html) this afternoon, which I think would be a good candidate as a readable serialization format. Presumably this has an MIT implementation in Squeak/Pharo that could be imported without much change.

PR #167, opened 2016-03-18
Limit tool search to classes if multiple classes match

jgfoster:
The SmalltalkSearchPresenter shows all the methods for the first class before showing any other classes. This limits the initial list of options to something more manageable. Since it does change the behavior I'm not sure if others will find it as useful as I do. What do you think? If you want to keep the existing behavior is it worth making this optional?

### DIFF
--- a/Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
+++ b/Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
@@ -154,14 +154,21 @@ rangeToCompleteAt: anInteger
 				ifFalse: [anInteger to: view caretPosition - 1]]!
 
 searchItemsStartingWith: aString maxItems: anInteger 
-	| result classes |
+	| classNamePrefix result classes |
 	aString first isLowerCase 
 		ifTrue: 
 			[| matchingSymbols |
 			matchingSymbols := self selectorsStartingWith: aString maxItems: anInteger.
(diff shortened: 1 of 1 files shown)


Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
```diff
@@ -154,14 +154,21 @@ rangeToCompleteAt: anInteger
...
 
 searchItemsStartingWith: aString maxItems: anInteger 
-	| result classes |
+	| classNamePrefix result classes |
 	aString first isLowerCase 
 		ifTrue: 
 			[| matchingSymbols |
 			matchingSymbols := self selectorsStartingWith: aString maxItems: anInteger.
 			^matchingSymbols collect: [:each | each displayString]].
+	classNamePrefix := aString upTo: $>.
```
blairmcg: Good idea, but how about something like the following, which also delays including metaclasses. Without that the "classes size > 1" condition below is going to be true even when there is really only one matching class.

```
	filter := (aString includes: $>) 
				ifTrue: 
					[| classNamePrefix |
					classNamePrefix := aString upTo: $>.
					[:each | each name asString = classNamePrefix]]
				ifFalse: 
					[(aString includes: Character space) 
						ifTrue: [[:each | each name beginsWith: aString]]
						ifFalse: [[:each | each isMetaclass not and: [each name beginsWith: aString]]]].
	classes := environment classes select: filter.
```

PR #187, opened 2016-04-06
CRTLibraryImportTest

jgfoster:
Help find missing exports.

### DIFF
--- a/Core/Object Arts/Dolphin/Lagoon/CRTLibraryImportTest.cls
+++ b/Core/Object Arts/Dolphin/Lagoon/CRTLibraryImportTest.cls
@@ -10,39 +10,23 @@ CRTLibraryImportTest comment: ''!
 !CRTLibraryImportTest categoriesForClass!SUnit! !
 !CRTLibraryImportTest methodsFor!
 
-defFilePaths
-	^#('Core\DolphinVM\ConsoleToGo\ConsoleToGo.def' 'Core\DolphinVM\InProcToGo\IPToGo.def' 'Core\DolphinVM\ToGoStub\togo.def')!
+stubNames
+	^#('ConsoleToGo.exe' 'GUIToGo.exe' 'IPDolphinToGo.dll')!
 
 test
 	"Ensure that .def files include proper CRTLibrary imports (#183)."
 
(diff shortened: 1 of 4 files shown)


jgfoster:
This should wait for https://github.com/dolphinsmalltalk/DolphinVM/pull/67 before it is merged.

Core/Object Arts/Dolphin/Lagoon/CRTLibraryImportTest.cls
```diff
@@ -0,0 +1,48 @@
...
+	instanceVariableNames: ''
+	classVariableNames: ''
+	poolDictionaries: ''
+	classInstanceVariableNames: ''!
+CRTLibraryImportTest guid: (GUID fromString: '{A3836AC0-E09A-4F44-B33F-3F87F52F2098}')!
+CRTLibraryImportTest comment: ''!
+!CRTLibraryImportTest categoriesForClass!SUnit! !
+!CRTLibraryImportTest methodsFor!
+
+defFilePaths
```
blairmcg: It seems a good idea to test this, but the problem is that the VM code is not necessarily available, e.g. for the AppVeyor build. I don't think requiring that one have both repos is really desirable (if that were the case, we may as well merge them into one). Another way to do this will be to check the actual exports of the togo stubs. This is easier than it might at first appear:

> f := FileLocator installRelative localFileSpecFor: 'GuiToGo.exe'.
> l := ExternalLibrary open: f.
> procs := (CRTLibrary selectMethods: [:each | each isKindOf: ExternalMethod]) collect: [:each | each functionName].
> "#getProcAddress: will throw if proc not found, or use getProcAddress:ifAbsent:"
> missing := procs select: [:each | (l getProcAddress: each ifAbsent: []) isNil]
> 
> "-> an IdentitySet('_close' '_finite' 'freopen' '_fpclass')"

blairmcg:
@jgfoster, I've committed your change to the DolphinVM repo and pushed a new tag, so version 7.0.11 is ready to consume for this submission once the test is fixed as described above.

jgfoster:
Thanks, I'll work on the suggested change...

Issue #209, opened 2016-06-08
Trigger an event when a package is installed


Sergiomdf:
It'd be great if the package manager could trigger an event everytime a package is installed, for a build process to log the "progress" of package installation.

blairmcg:
Seems reasonable

PR #210, opened 2016-06-09
Trigger an event when a package is installed (#209)

Sergiomdf:


Core/Object Arts/Dolphin/Base/PackageManager.cls
```diff
@@ -510,7 +510,10 @@ install: aString
 	[newPackages do: 
 			[:each | 
 			self addPackage: each.
-			[self basicInstall: each] ifCurtailed: [self basicUninstall: each].
+			[
+				self basicInstall: each.
+				self trigger: #packageInstalled with: each.
```
blairmcg: The event selector should be #packageInstalled: (i.e. with a terminating colon) because it has an argument. Although this is mainly a convention, it is standard practice for reasons of readability.

blairmcg:
Sergio I left a comment on the changed file to be addressed.

Thanks for creating the issue and associating with the PR. Going forward we want all PRs to relate to an issue.

Core/Object Arts/Dolphin/Base/PackageManager.cls
blairmcg: I did also wonder if this was the right place to trigger the event, given that only the basic installation has been performed and the manager is not yet observing the package. However it makes since in that if an event handler throws an exception, the package will be uninstalled by the curtailment handler. Therefore I think this is right.

Issue #214, opened 2016-06-23
Float >> #printOn:decimalPlaces: ignores decimal places when sending to 0.0


Sergiomdf:
When `#printOn:decimalPlaces:` is sent to `0.0`, the decimal places argument (anInteger) is ignored. It always output '0.0'.

PR #215, opened 2016-06-23
Float >> #printOn:decimalPlaces: ignores decimal places when sending to 0.0 (#214)

Sergiomdf:


Core/Object Arts/Dolphin/Base/Float.cls
```diff
@@ -638,27 +638,25 @@ printOn: aStream base: anInteger
...
-			[self sign = -1 ifTrue: [aStream nextPutAll: '-'].
-			self = 0.0
-				ifTrue: [aStream nextPutAll: '0.0']
-				ifFalse: 
-					[self
-						absPrintExactlyOn: aStream
-						base: 10
-						decimalPlaces: anInteger
-						showTrailingFractionalZeros: true]]!
+	self sign = -1 ifTrue: [aStream nextPutAll: '-'].
```
blairmcg: Please format this to use standard layout (Press Ctrl+W to auto-format it)

Core/Object Arts/Dolphin/Base/FloatTest.cls
```diff
@@ -361,7 +361,18 @@ testPrintOnDecimalPlaces
...
+
+	"Previous implementation was failing when sending to zero."
+	stream := String writeStream.
+	0.0 printOn: stream decimalPlaces: 0.
+	self assert: stream contents equals: '0'.
+	stream := String writeStream.
+	0.0 printOn: stream decimalPlaces: 1.
+	self assert: stream contents equals: '0.0'.
+	stream := String writeStream.
+	0.0 printOn: stream decimalPlaces: 2.
```
blairmcg: Thanks for updating the unit tests. 
At least one test case for -0.0 (negative zero) would be useful. I think your code will work, but it is a useful regression test.

blairmcg:
@Sergiomdf, I've made some minor comments on your PR. I'm happy to commit this when these are addressed.

PR #221, opened 2016-07-05
Changes for demo

objectarts:
Changes for a demonstration .. to be discarded


objectarts:
I don't think this should go into the master

PR #244, opened 2016-10-08
Eliminate rogue ".img" file in early boot process

daniels220:
Seems like this was mentioned many months ago as a silly minor annoyance... Poking around SessionManager, it seems like getting rid of it is really simple, so unless I'm missing something, here you go.

### DIFF
--- a/Boot.st
+++ b/Boot.st
@@ -83,8 +83,7 @@ devsesh onPreSaveImage.
 	maxObjects: 24576*1024
 ] ensure: [devsesh onPostSaveImage]!
 
-"Remove rogue .img and unnecessary .chg files"
-File delete: (File default: SessionManager current imagePath extension: 'img').
+"Remove unnecessary .chg files"
 File delete: 'DBOOT.chg'!
 
 SessionManager current quit!
--- a/Core/Object Arts/Dolphin/Base/BootSessionManager.cls
+++ b/Core/Object Arts/Dolphin/Base/BootSessionManager.cls
@@ -292,6 +292,6 @@ updateBootImage
 imageExtension
 	"Answer the suffix for an executable image file"
 
-	^'img'! !
+	^'img7'.! !
 !BootSessionManager class categoriesFor: #imageExtension!constants!public! !
 


Core/Object Arts/Dolphin/Base/BootSessionManager.cls
```diff
@@ -292,6 +292,6 @@ updateBootImage
 imageExtension
 	"Answer the suffix for an executable image file"
 
-	^'img'! !
+	^'img7'.! !
```
blairmcg: If this is required it should really be copied into the PreBoot.St script as it belongs to the BootSessionManager that performs the boot.

Boot.st
```diff
@@ -83,8 +83,7 @@ devsesh onPreSaveImage.
 	maxObjects: 24576*1024
 ] ensure: [devsesh onPostSaveImage]!
 
-"Remove rogue .img and unnecessary .chg files"
-File delete: (File default: SessionManager current imagePath extension: 'img').
```
blairmcg: This doesn't seem quite right. "SessionManager current" will be an instance of the DevelopmentSessionManager at this point.
Do we know what is causing this "rogue" image to be saved in the first place?

Core/Object Arts/Dolphin/Base/BootSessionManager.cls
daniels220: True, but the very next thing Boot.st does after loading PreBoot.st is update the BootSessionManager from its .cls file, so having this there is more than early enough. The critical point, as per my other comment, is Boot.st line 58, the first time the image is saved.

Boot.st
daniels220: The cause of the rogue .img is the incorrect return value of `BootSessionManager class>>imageExtension`:
* Boot.st line 58 saves the image to its intended new filename just before invoking DolphinProduct>>boot.
* Then see BootSessionManager>>saveImageTo:, where (after updating the value of `imagePath`) it calls `self snapshot: self imageFileName`.
* #imageFileName constructs its answer from the answers to #imagePath and #imageExtension, and #imageExtension simply forwards class-side.

As far as `SessionManager current` being a DevelopmentSessionManager by this point in the boot process: This is true, but all the script cared about at that point is that it had the correct value for `imagePath`, which of course it did, having copied it from the BootSessionManager as part of #installNew.

blairmcg: Sorry I misread the diff when I looked at this before. I overlooked all the visual signals and thought you were adding this line to the script. That's fine then.

PR #245, opened 2016-10-08
Remove redundant and subtly-wrong ListView>>selectionByIndex:ifAbsent:.

daniels220:
I stumbled across this while writing a test, doing something like:

```
someListView selectionsByIndex: #(1 2 4).
"Do some other stuff..."
someListView selectionByIndex: #(1).
"Make some assertions about how other views change as a result"
```

And being baffled why those assertions were failing—then looking at the view and realizing that the selection never changed. I traced the cause to a bug in the way ListView>>selectionByIndex: decides whether or not it can just bail because nothing is changing, such that trying to set the selection to exactly one of the (multiple) items that are already selected will bail without deselecting the others.

It seems like the inherited implementation already forwards to #selection*s*ByIndex: in a reasonable way (which works fine for a view in single-select mode), so the override can simply be removed.


blairmcg:
I have a test that fails on the existing code, so I'll merge this...

PR #246, opened 2016-10-10
Add test for ListView>>selectionByIndex:ifAbsent: bug

blairmcg:


blairmcg:
Realised I've reduced the coverage of listboxes in multi-select mode. Abandoning this for now.

PR #247, opened 2016-10-11
Replace deprecated SQLError function with SQLGetDiagRec

daniels220:
I was investigating an error thrown by a database operation that Dolphin couldn't figure out what it actually meant, and fell back to simply "SQL_ERROR" as the description. Looking at the docs, the SQLError() function used to retrieve the error description is deprecated. Lo and behold, replacing it with the successor SQLGetDiagRec() revealed the problem.

I also noticed and fixed a couple places where the Smalltalk selector reflected the current name of an ODBC function, but the actual declared name in the body was a slightly-different, older, deprecated name. (Two places where a function name used to end with "Options" and now ends with "Attr", with the same signature and behavior.)

### DIFF
--- a/Core/Object Arts/Dolphin/Database/DBConnection.cls
+++ b/Core/Object Arts/Dolphin/Database/DBConnection.cls
@@ -1447,8 +1447,7 @@ transact: aDBConnection action: anInteger
 				sqlEndTran: SQL_HANDLE_DBC
 				handle: aDBConnection asParameter
 				completionType: anInteger.
-	ret ~= SQL_SUCCESS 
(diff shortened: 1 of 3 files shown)


daniels220:
Yes, I've noticed that this is one of if not the only package that is unchanged between Dolphin 6.0 and 7.

I looked at a couple lists of deprecated functions with direct equivalents that are auto-mapped by the driver system, and those are now all replaced by their modern equivalents. I wasn't able to find a straight-up list of all deprecated functions, but there might well not be more, in which case this is it.

PR #248, opened 2016-10-12
Avoid loading MVP code in the boot script if the product to boot was passed as a parameter

daniels220:
I remember some discussion on the newsgroup some months ago about the advantages of deploying applications by "building them up" from the boot image rather than "tearing them down" from a development image with Lagoon. I noticed shortly after using Dolphin 7 for the first time that the boot script loads MVP code early on even if it does not need to display a ChoicePrompter because the product to boot was provided as a command-line argument. It seems like eliminating this requirement would make it easier to boot a command-line app, or even just a simple MVP app that doesn't use ChoicePresenter/Prompter.

Dolphin Products itself also depended on Dolphin MVP Base due to using class `RGB` for `#defaultSystemFolderColor`. Since the call to that is guarded behind a `Smalltalk at: #SmalltalkSystemShell ifPresent:`, it seemed reasonable to convert this to a `Smalltalk at:` also in order to break that dependency.

Perhaps ideally, `DolphinProduct`, with all its infrastructure for the boot process, would live in its own package, with `DolphinCoreProduct` and subclasses in another package. It could then override `#boot` to add `#installSystemFolderOptions` and `#installSplash`, allowing them to be left out of the base class entirely and keeping the dependency graph reasonable without indirection. If this seems like a good idea to you, I'm happy to go ahead and do it, either on this branch or as a separate pull request.

### DIFF
--- a/Core/Object Arts/Dolphin/Installation Manager/DolphinCoreProduct.cls
+++ b/Core/Object Arts/Dolphin/Installation Manager/DolphinCoreProduct.cls
@@ -100,7 +100,13 @@ contents
 	^contents!
 
 defaultSystemFolderColor
-	^RGB red: 41 green: 124 blue: 182!
+	"Reference RGB indirectly to eliminate hard dependency on Dolphin MVP Base.
+	This method will only be called if SmalltalkSystemShell is loaded, in which case RGB will be as well."
+
+	^(Smalltalk at: #RGB)
+		red: 41
+		green: 124
+		blue: 182.!
 
 fullName
 	"Answer the full product name associated with the receiver"
--- a/Boot.st
+++ b/Boot.st
@@ -18,27 +18,20 @@ Package manager install: 'Core\Object Arts\Dolphin\Registry\Dolphin Registry Acc
 "We need this to install any packages containing binaries"
 Package manager install: 'Core\Object Arts\Dolphin\System\Base64\Dolphin Base64.pax'!
 
-"We'll need MVP bits in order to be able to display the choice prompter for the image version to boot"
-Package manager 	install: 'Core\Object Arts\Dolphin\Base\Dolphin Base.pax'!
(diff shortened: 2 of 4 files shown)


blairmcg:
Personally I'm not convinced that constructing an image by loading packages will result in a material improvement. Yes, it is a much better approach than the traditional Smalltalk image stripping process, but that is not the primary mechanism by which Lagoon works. Lagoon starts by uninstalling the packages that are not part of the dependency graph, so (assuming the packages uninstall cleanly, which is generally not that difficult to arrange) the net result will be little different that an additive approach. Lagoon also (optionally) does more traditional image stripping after that, which, after removing unecessary packages, results in significant further trimming. Ironically I think if you just did additive construction based on packages you could end up with much more stuff in the result than Lagoon can achieve. Furthermore with both approaches you have to ensure that the package dependency graph is accurate or you will not get a valid application, so there is no particular "development rigour" advantage to the additive approach in terms of requiring cleanly defined packages with clear dependencies.
That doesn't mean that what you suggest is not worth doing - if you think it would be interesting, by all means go ahead.

Core/Object Arts/Dolphin/Installation Manager/DolphinCoreProduct.cls
```diff
@@ -100,7 +100,13 @@ contents
 	^contents!
 
 defaultSystemFolderColor
-	^RGB red: 41 green: 124 blue: 182!
+	"Reference RGB indirectly to eliminate hard dependency on Dolphin MVP Base.
```
blairmcg: Another way to do this would be to just return an array of the R-G-B values here and use that to construct an RGB in the caller, but I don't think it is worth changing.

daniels220:
Honestly my main motivation here was cleaning up the boot script—I'm not personally that interested in additive deployment, but it seemed worth paying attention to while I was in the area, for the sake of others who were.

PR #254, opened 2016-11-09
Optimize Package>>allPrerequisites, #hasCyclicPrerequisites, and #hasUncommittedPrerequisites

daniels220:
The current implementation has a roughly-exponential worst-case, which with large numbers of packages means that turning on the status icons in the package browser, or even just attempting to uninstall a package (which calls #hasCyclicPrerequisites), freezes the image for well upwards of a minute.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/Package.cls
+++ b/Core/Object Arts/Dolphin/Base/Package.cls
@@ -882,13 +882,9 @@ hasImageStripper
 hasUncommittedPrerequisites
 	"Private - Answer whether any of the receivers prerequisites or their
 	prerequisites are current uncomitted. If so the package should not be saved
-	as it will not be reloadable into an image without the uncommitted objects.
-	N.B. This method will go infinitely recursive if the package has cyclic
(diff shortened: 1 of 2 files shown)


Core/Object Arts/Dolphin/Base/Package.cls
```diff
@@ -42,6 +42,15 @@ aboutBlockBytes: aByteArray
 
 !
 
+addAllPrerequisitesTo: aSet
+	"Private - Add to <aSet> all those Packages which must be loaded in
+	before the receiver may be loaded in."
+
+	self prerequisites
+		do: [:each | (aSet includes: each) ifFalse:
+					[aSet add: each.
```
blairmcg: Double lookup (once to check inclusion and once to add) can be avoided here by using Set>>addNewElement:. This adds the element if not already present returning true if it was added, or false if already in the set.

daniels220: Right, I'm not used to that yet (since it wasn't there in D6). Thanks for the reminder.

PR #255, opened 2016-11-14
KernelLibrary class>>#’isWine’ doesn’t work properly in runtime

jgfoster:
It checks to see if NTLibrary is present, but by default it is stripped. This causes it to be included. An alternate edit would be to change the KernelLibrary method. Comments and suggestions welcome!

### DIFF
--- a/Core/Object Arts/Dolphin/Lagoon/ImageStripper.cls
+++ b/Core/Object Arts/Dolphin/Lagoon/ImageStripper.cls
@@ -1242,6 +1242,7 @@ requiredClasses
 	^(IdentitySet new: 32)
 		add: Process;
 		add: ProcessorScheduler;
+		add: NTLibrary;		"see KernalLibrary class>>#'isWine' "
 		add: self runtimeSessionManagerClass asSessionManagerClass;
 		addAll: self vmReferencedClasses;
 		yourself!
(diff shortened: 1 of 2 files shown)


jgfoster:
Sounds good. I'll need to research how to put something into the Dolphin package. As I recall, at the beginning it was more complicated to modify. (In fact, everything in 'Dolphin Base' really should be in Dolphin but we didn't have a way to do that at first.)

blairmcg:
When I suggested that, I thought it would "just work". Adding stuff to the base is supposed to be easy since I modified the boot process to reload the system package before proceeding with package installation, however it looks like it is easy to add new methods/state to existing classes, but it doesn't actually load newly defined classes :-(. Let me look into that, as it is supposed to just work so you shouldn't have to invest a lot of effort.

blairmcg:
That's done. I moved NTLibrary to the Dolphin package to test it, so you can abandon this one if you like. With the change it is easy to add new BCL classes, although in general we should be looking to move stuff out, rather than adding more. I think the command line processing stuff justifies its own package, and I can't see a good reason to consolidate it into the BCL since there are no BCL dependencies on it. 'Dolphin Base' could be renamed/repurposed for that, since it now contains only the command line processing classes.

jgfoster:
Thanks!

PR #259, opened 2016-11-16
Changing the superclass and instanceVariables of a class in a single operation can produce apparent duplicate or missing instance variables

daniels220:
For example:

```smalltalk
"Given this state:"
Object subclass: #ClassA
    instanceVariableNames: 'var1'
    classVariableNames: ''
    poolDictionaries: ''.
Object subclass: #ClassB
    instanceVariableNames: 'var1 var2'
    classVariableNames: ''
    poolDictionaries: ''.
"Then, presumably later:"
ClassA subclass: #ClassB
    instanceVariableNames: 'var2'
    classVariableNames: ''
    poolDictionaries: ''.
```

The mutation can be performed in-situ, and #mutateInSitu only checks if the superclass has changed—presumably assuming that an instance variable change would require a full mutation, which it usually would, but not in this one unique case.

So, the current implementation does not set the instance variables of `ClassB`, and both classes end up declaring `var1`. `ClassB`'s `#instSize` is correct (2), but any new methods referencing `var2` will probably get a BoundsError when run, because the Compiler thinks `var2` is at index 3, instead of the correct 2.

This also happens in reverse—given `ClassB` is already a subclass of `ClassA`, evaluating `Object subclass: #ClassB instanceVariableNames: 'var1 var2'` ...etc. leaves ClassB with an `#instSize` of 2, but thinking it only has `var2`, so methods which reference `var1` will presumably not compile. (In both cases, no recompilation is triggered, so existing methods will continue to work correctly until recompiled.)

One note: I added `self recompilationRequired: true` to match the superclassIsBeingChanged case, but I'm not sure why recompilation would ever be required when mutating in-situ—doesn't the requirement that #allInstVarNames remains unchanged automatically mean that all existing methods are valid?

### DIFF
--- a/Core/Object Arts/Dolphin/Base/ClassBuilder.cls
+++ b/Core/Object Arts/Dolphin/Base/ClassBuilder.cls
@@ -488,10 +488,15 @@ mutateInSitu
 	"Private - Mutate the class being modified without mutating its instances.
 	In fact, we only need to change the superclass."
 
-	self superclassIsBeingChanged ifFalse: [^self].
-	currentClass class setSuperclass: (self class classOf: superclass).
-	currentClass setSuperclass: superclass.
-	self recompilationRequired: true!
+	self superclassIsBeingChanged
+		ifTrue:
+			[currentClass class setSuperclass: (self class classOf: superclass).
+			currentClass setSuperclass: superclass.
+			self recompilationRequired: true].
+	self instanceVariablesAreBeingChanged
+		ifTrue:
+			[self setInstanceVariablesOf: currentClass.
+			self recompilationRequired: true].!
 
 mutateToNewClass
 	"Private - Set currentClass to a mutation of itself based on the information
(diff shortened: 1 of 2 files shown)


blairmcg:
"doesn't the requirement that #allInstVarNames remains unchanged automatically mean that all existing methods are valid" - as far as the inst var slots, yes, but changing the superclass can change other aspects of the compilation scope, e.g. available Class Variables. ## expressions, which are effectively evaluated in the context of the class, might also evaluate differently, e.g. ##(self).

Core/Object Arts/Dolphin/Base/ClassBuilder.cls
```diff
@@ -488,10 +488,15 @@ mutateInSitu
 	"Private - Mutate the class being modified without mutating its instances.
 	In fact, we only need to change the superclass."
```
blairmcg: This comment was a correct description of the actual behaviour, but not the desired behavior. With your change it no longer describes either :-)

daniels220:
Riiiight, that makes sense why recompilation would be required when changing superclass. In turn I realize it's not possible to reach this point with inst-var changes and no superclass change—by definition this means the overall inst var layout _would_ change and we would not be doing the mutation in-situ. So, the second `self recompilationRequired: true` would appear to be entirely redundant—does this sound right?

Also, re: the comment—d'oh! In general, would you like me to fix issues like this, or does your merging the PR mean either you plan to do it, or don't think it's that big of a deal? (I'm happy to fix it, and I guess if you don't plan to do it, I probably will—the comments in Dolphin are usually both accurate and actually helpful, and I'd hate to knowingly introduce an inaccuracy...)

PR #261, opened 2016-11-22
Delegate choosing the icon for a CompiledMethod to the development system

daniels220:
Allows for customization of method icons without further editing of base packages (by subclassing SmalltalkSystem). Examples: Different icons for auto-generated methods, basic accessors, methods marked subclassResponsibility/shouldNotImplement/notYetImplemented/etc.

Fall back to simply the class-side default (the Public icon, effectively) if no development system is present. This seems reasonable, in that I would not expect method icons to be displayed as part of a deployed executable, or if they are it seems reasonable for the developer to need to take this into account in the image-stripping process.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/CompiledMethod.cls
+++ b/Core/Object Arts/Dolphin/Base/CompiledMethod.cls
@@ -1,4 +1,4 @@
-"Filed out from Dolphin Smalltalk X6"!
+"Filed out from Dolphin Smalltalk 7"!
 
 CompiledCode variableSubclass: #CompiledMethod
 	instanceVariableNames: ''
@@ -77,12 +77,9 @@ hasChanged
 icon
 	"Answers an Icon that can be used to represent this object."
 
-	self isDeprecated ifTrue: [^self class deprecatedIcon].
-
-	^self isPrivate
-		ifTrue: [self class privateIcon]
-		ifFalse: [self class publicIcon]
-!
+	^(Smalltalk at: #SmalltalkSystem ifPresent: [:a | a current])
+		ifNil: [super icon]
+		ifNotNil: [:a | a iconForMethod: self]!
 
 infoTip
 	"Private - Answer a suitable 'info tip' for the receiver."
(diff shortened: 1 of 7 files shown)


Core/Object Arts/Dolphin/IDE/Base/SmalltalkSystem.cls
```diff
@@ -1888,6 +1888,12 @@ hierarchyBrowserClass: toolClass
 
 	hierarchyBrowserClass := toolClass!
 
+iconForMethod: aCompiledMethod
+	"Answer an icon that can be used to represent the <CompiledMethod> argument in a development environment."
+
+	aCompiledMethod isDeprecated ifTrue: [^CompiledMethod deprecatedIcon].
```
blairmcg: I think you should move the CompiledMethod class>>deprecatedIcon|privateIcon|publicIcon methods to SmalltalkSystem too. They are private, and there isn't really a good reason to override these in the CompiledMethod class hierarchy. I realise this will mean replacing the refs in StsMethodEdition, but it is already dependent on the Development System package.

Core/Object Arts/Dolphin/IDE/Base/Development System.pax
```diff
@@ -1982,13 +1983,21 @@ browse
     
     	^Smalltalk developmentSystem browseMethod: self!
 
+isDeprecated
```
blairmcg: I see you've moved this method to the development system package (which is where it should have been in the first place), but it may as well become SmalltalkSystem>>isDeprecatedMethod:. Even though it was public before I don't think anyone will have called it.

Core/Object Arts/Dolphin/Base/CompiledMethod.cls
```diff
@@ -77,12 +77,9 @@ hasChanged
 icon
 	"Answers an Icon that can be used to represent this object."
 
-	self isDeprecated ifTrue: [^self class deprecatedIcon].
-
-	^self isPrivate
-		ifTrue: [self class privateIcon]
-		ifFalse: [self class publicIcon]
-!
+	^(Smalltalk at: #SmalltalkSystem ifPresent: [:a | a current])
```
blairmcg: Since this super-sends when the development system is not present, a simpler solution would be to make it a loose method in the devsystem package, and then it can just be `^SmalltalkSystem current iconForMethod: self`

daniels220:
Those are all excellent ideas, which I have now implemented. Thanks for the feedback!

PR #262, opened 2016-11-22
Ensure that copying a parse node fully copies the tree below it.

daniels220:
This was clearly intended, as most nodes implement #postCopy in this fashion, but a few did not do so correctly or at all.


daniels220:
Unfortunately Pharo has nothing new to add—they check lots of different trees for equality-after-copy, but don't in any way attempt to verify that the copy is truly independent. Thanks for nudging me to look at Pharo again, though—it looks like they've made some improvements in the general area of the parser that would be cool to bring to Dolphin. I've created #265 to track this separately.

In starting to write such a test myself, I noticed that SmalltalkParserTest and ParserTest have at least 32 selectors in common for which the source code is equivalent (with very minor deviations, e.g. substituting `self parserClass` for a direct reference to `RBParser`). Many of them have to do with searches and rewrites—things that the Smalltalk Parser package alone is not capable of. It strikes me that the separation between "Smalltalk Parser" and "RBParser" may be a holdover from needing to keep RBParser out of the Community Edition image, while preserving basic parsing (and formatting) capabilities. Might it now make sense to merge everything onto ParserTest?

blairmcg:
If I recall correctly a copy of ParserTest was made primarily because we wanted to keep the RB port for Dolphin as close as possible to the original code, with only clearly marked modifications where absolutely necessary. This was to ease consumption of future updates to the code base. 
It probably would make sense to rationalise this now to reduce the duplication.

daniels220:
Not sure if Github will have notified you about changes (rebase shenanigans)... One note: I left alone the few tests that existed on SmalltalkParserTest and not on ParserTest—seemed reasonable since they test functionality added/changed for Dolphin.

Core/Contributions/Refactory/Refactoring Browser/Tests/ParserTest.cls
```diff
@@ -693,13 +700,34 @@ verifySelector: aSelector inClass: aClass
...
+				isEqualTo: (Compiler compile: source in: aClass))!
+
+verifyTree: copiedNode isDeepCopyOf: originalNode
+	self assert: originalNode class == copiedNode class.
+	self deny: originalNode == copiedNode.
+	self assert: originalNode = copiedNode.
+	originalNode children with: copiedNode children
+		do: 
+			[:originalChild :copiedChild |
+			"Messages in a cascade share their receiver coming out of the parser, but #copy gives them each a separate copy."
```
blairmcg: That seems like another bug in the RB parse node copy operation.

Issue #294, opened 2017-01-20
ByteArray>>itcAsInteger uses wrong endian order


bernhardkohlhaas:
The implementation of ByteArray>>itcAsInteger treats the first byte as MSB, however I believe that the first Byte should be the LSB, so that the result is equal to 'wordAtOffset: 0' or 'dwordAtOffset: 0' (for array sizes 2 and 4 respectively).

Therefore I suggest that the iteration should be done in reverse order.

Proposed implementation:

    ByteArray>>itcAsInteger

    	"Since the LSB is at index 1, we need to reverse iterate over the receiver"

    	| answer |
    	answer := 0.
    	self reverseDo: [:each | answer := answer * 256 + each].
    	^answer

blairmcg:
As discussed in the PR, this may be because it is designed for portability across platforms.

PR #295, opened 2017-01-20
ByteArray>>itcAsInteger uses wrong endian order

bernhardkohlhaas:
This commit fixes issue #294 

P.S.: My first pull request on github. Please let me know, if it needs to be done differently

### DIFF
--- a/Core/Contributions/ITC Gorisek/Dialect Abstraction Layer.pax
+++ b/Core/Contributions/ITC Gorisek/Dialect Abstraction Layer.pax
@@ -373,7 +373,12 @@ itcAsHexString
 	^stream contents!
 
 itcAsInteger
-	^self inject: 0 into: [:sum :each | sum * 256 + each]! !
+	"Since the LSB is at index 1, we need to reverse iterate over the receiver"
+
+	| answer |
+	answer := 0.
+	self reverseDo: [:each | answer := answer * 256 + each].
+	^answer! !
 !ByteArray categoriesFor: #itcAsHexString!accessing!public! !
 !ByteArray categoriesFor: #itcAsInteger!accessing!public! !
 


bernhardkohlhaas:
P.P.S.: I didn't include all the other changed files from the package save, since the only difference was the "Filed out from" message now refers to Dolphin Smalltalk 7 instead of Dolphin Smalltalk X6.2.

I hope that is the correct procedure.

blairmcg:
@bernhardkohlhaas, I don't know if this is correct or not. Although you are right that if is supposed to be converting from the standard x86 format binary representation of an integer it should be little-endian, this method is part of David Gorisek's dialect abstraction layer, and it may be deliberately converting from a normalised form where the choice is big endian to enable interop between machines with different byte orders. It's difficult to say, however, as the only senders of #itcAsInteger in the image look to me to be sending to strings.

bernhardkohlhaas:
@blairmcg, you raise a very valid point. Do you have any way to check with David?

Or alternatively would it make sense to add a method ByteArray>>asInteger in the Dolphin package with the proposed code?

bernhardkohlhaas:
On second thought instead of ByteArray>>asInteger something like this might be more in the spirit of the ByteArray class:

    ByteArray>>integerOfByteSize: byteSizeInteger atOffset: ofsInteger
	    "Answer an unsigned integer with <byteSizeInteger> bytes 
	    at offset (i.e. zero relative) <ofsInteger> IN the receiver."

	    | answer |
	    answer := 0.
	    ofsInteger + byteSizeInteger to: ofsInteger + 1
	    	by: -1
	    	do: [:index | answer := answer * 256 + (self at: index)].
	    ^ answer

blairmc-msft:
Do you have a need for such a method Bernhard? 
I'm not aware of a general purpose need. There are existing methods to efficiently read various sizes of signed/unsigned integers from a ByteArray at specified offsets, assuming little-endian representation, already. Most of these have primitive support as well.

bernhardkohlhaas:
Blair, you are right that the existing methods are more efficient, but they require you to know the size of the integer at compile time. I'm writing some code to access the MS Crypto API for a Random Number Generator and there is is useful to take the size of the integer as a parameter to effectively generate random numbers in different ranges.

Indeed my implementation above is much less efficient, compared to the existing xxxAtOffset: methods, e.g. 5-9 times better for dwordAtOffset: depending on whether the result is still in the SmallInteger range or not. FWIW I changed my method implementation (using a precompiled Array with BlockClosures as a jump table, depending on array size) that brought the penalty down to 2.5-6 times, but that's still significant.

For  my particular use case I'm now considering a different approach, overallocating the ByteArray and working with multiples of 4 as a size, leaving the final bytes zero and then always using dwordAtOffset: 

In any case, if you don't think the method above (even with a better implementation) is suitable for the Core Image, then feel free to close this pull request and the associated issue #295 I'm happy to keep it in one of my own packages.

Best,
Bernhard

blairmcg:
Bernhard, Ok I understand your requirement. I'm not sure precisely which API you are calling, but if it is [CryptGenRandom](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379942(v=vs.140).aspx) then passing a newly instantiated LargeInteger of appropriately sized (i.e. round up to 4, and add 4 to ensure positive) to the function is your most efficient option. This will work well because Dolphin's large integers use the native Intel 2's compliment representation - all you need do then is normalize the result to ensure it is the minimum representation.
I'd be tempted to add a wrapper method to your package with a body something like this:

```
	| random |
	random := LargeInteger new: ((randomSize roundUpTo: 4) bitShift: -2) + 1.
	(self
		cryptGenRandom: hProvider
		dwLen: anInteger
		pbBuffer: random yourAddress) ifFalse: [Win32Error signal].
	^random normalize
```
where randomSize is the length in bytes of random integer you want, and hProvider is the cryptographic provider handle from CryptAcquireContext.

I think this belongs in a cryptographic extension, rather than the base package.

HTH

Blair

bernhardkohlhaas:
@blairmcg ,

Thank you so much for taking the time and giving such a thorough response, I really appreciate it. 

Yes, your suggestion is exactly what I'm looking for and I'm using that very API as well. Now I've learned something new about using LargeIntegers. :)

Thanks again & best regards,
Bernhard

PR #296, opened 2017-01-21
Unpackage STBPrefixConstants in PreBoot

daniels220:
Otherwise the package still thinks it's there (it is still listed in `globalNames`), when of course it has actually been removed. Since the original removal was done in PreBoot.st, it seemed reasonable to do this cleanup there also.


blairmc-msft:
LGTM

PR #297, opened 2017-01-21
Fix "Array should not implement #remove:ifAbsent:" error during image-stripping process

daniels220:
Seemed reasonable to just switch to using #difference: instead.

Is there a reasonable way to write a test for this, or due to the nature of the image-stripping process is it too difficult? Would it make sense to add stripping one of the example applications to the AppVeyor build as a substitute for SUnit tests?


blairmcg:
I came across this myself yesterday, and fixed it the same way.

PR #298, opened 2017-01-21
Fix nil DNU #identityIncludes: error adding classes to the BCL

daniels220:
The change to PreBoot.st in [this commit](https://github.com/dolphinsmalltalk/Dolphin/commit/282be5fd22a0040ef29f8678ab9ea0dc8b78a62a) seems incomplete—until the new implementations of `#isImportedClass:` are loaded when ClassLocator.cls and Package.cls are filed in, trying to package a new BCL class hits a nil where a collection was previously expected. Seems reasonable to just update those methods in PreBoot.st along with nilling the variable.


blairmcg:
LGTM

PR #305, opened 2017-02-07
Don't complete keyword-selector symbol literals as if they were keyword message sends

daniels220:
Decide whether to perform keyword completion based on the view's style rather than the presence or absence of a leading $#. With the addition of images to autocompletion lists, there will never be one, though this has always been an issue when completing keyword symbol literals using ST80 syntax inside a literal array.


blairmcg:
Glad you fixed this. It's been annoying me as well since I broke it.

PR #309, opened 2017-02-08
Show current product info on SmalltalkSystemShell

bernhardkohlhaas:
The VM info displayed in the title of SmalltalkSystemShell is not the current VM (but probably the one that the boot image was created with). Here an example for an image booted with the latest VM:

image

This fix forces a redetermination of the product details on system startup and adds a small adjustment in VMLibrary to provide (once again?) the productVersionString:

image

### DIFF
--- a/Core/Object Arts/Dolphin/Base/VMLibrary.cls
+++ b/Core/Object Arts/Dolphin/Base/VMLibrary.cls
@@ -141,7 +141,7 @@ defaultProductDetails
 		at: 5 put: self basePatchLevel;
 		at: 6 put: 'D7';
 		at: 7 put: nil; "Serial #"
-		at: 8 put: nil; "Boot info"
+		at: 8 put: version productVersionString; "Boot info"
 		yourself!
 
 dump: msgString path: pathString stackDepth: slotsInteger walkbackDepth: framesInteger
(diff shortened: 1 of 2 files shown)


bernhardkohlhaas:
After 2nd (and 3rd) thoughts, perhaps it is intentional that the version of the boot image is shown, rather than the version of the current VM. The comment in VMLibrary>>defaultProductDetails for array slot 8 seems to indicate as a much.

However there isn't really any way to properly retrieve the boot image info in VMLibrary (which is perhaps, why that slot was set to nil).

If this warrants further discussion as to what _should_ happen, I'm happy to withdraw this pull request.

blairmcg:
Yes, it is intended to show the version of the sources the image was booted against. The version shown is the 'git describe' of the image repo - it's not related to the VM version - I realise that is confusing. Part of the issue is that the VM and image are in separate repo's, but often need to be modified in sync. Although there are good reasons for this, it makes managing the versions more painful, and I'm afraid I haven't been very good at re-labelling the image repo when I make a breaking change to the VM repo.

bernhardkohlhaas:
Thanks for the explanation, Blair. I'll close this pull request then.

PR #327, opened 2017-03-04
Fix ListView #selectionChanging: event to have the correct #newSelections when the view is in multi-select mode

daniels220:
I converted a formerly single-select list to multi-select and noticed that, when making a multiple selection, the proposed `newSelections` of the `#selectionChanging:` event didn't at all reflect the selection(s) that the view would actually end up with. I spent some time playing with the view, learning its actual behavior, and came up with this implementation which I _think_ is correct.

It seems reasonable for something like this to have tests, but I'm not sure how best to do this. I played around some with posting "fake" WM_MOUSEDOWN/WM_MOUSEUP messages to the view and was able to automatically "play through" some scenarios. Does this approach make sense to you? If so, are there any other tests that do this, that might establish conventions I could follow? Or if not, do you have another approach altogether that you would suggest?

### DIFF
--- a/Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
+++ b/Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
@@ -1473,6 +1473,45 @@ moreParamsAt: n put: anObject
 	self moreParams at: n.
 	^self moreParams at: n put: anObject!
 
+newSelectionsFromEvent: event
+	"Private - Answer the <integer> selection that would occur if the <MouseEvent>,
+	event, was was passed to the control."
+
+	| itemClicked anyKeysDown newSelections markIndex markRange markIsSelected |
+	itemClicked := self itemFromPoint: event position.
+	self isMultiSelect ifFalse: [^itemClicked ifNil: [#()] ifNotNil: [:item | Array with: item]].
+	newSelections := self selectionsByIndex.
+	""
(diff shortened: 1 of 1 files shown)


Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
```diff
@@ -1473,6 +1473,45 @@ moreParamsAt: n put: anObject
...
+	markIndex := self selectionMarkIndex.
+	markRange := markIndex == 0
+				ifTrue: [itemClicked to: itemClicked]
+				ifFalse: [(markIndex min: itemClicked) to: (markIndex max: itemClicked)].
+	"Only Shift"
+	event isCtrlDown ifFalse: [^markRange asArray].
+	"Ctrl and Shift"
+	newSelections := newSelections asSet.
+	markIsSelected := newSelections includes: markIndex.
+	markRange do: (markIsSelected
```
blairmcg: Set operations, `union:` and `distinct:`?

blairmcg:
I'm going to merge this now - we can worry about the tests later, after all this was completely broken before.

daniels220:
Makes sense, thanks. The testing strategy you suggested sounds good, and I was/am planning to write some tests, I just haven't had time so far. Thanks also for your feedback above—I'll incorporate it as I'm dealing with tests.

PR #341, opened 2017-03-14
Don’t strip ExternalStructure class>>#’offsetOf:’.

jgfoster:
A recent commit (619963bb3261408afc6a821ce8c540a94f7b9cc5) changed BasicImageStripper>>#prepareExternalStructuresNotifying: to remove #'offsetOf:' from ExternalStructure class. Unfortunately it is still sent by NMLVDISPINFO class>>#'itemFromNMHDR:' so we get a runtime error on certain operations.

{1A791118: cf 1A7910F1, sp 1A79112C, bp 1A791114, ip 4, NMLVDISPINFO class(Object)>>doesNotUnderstand:}
	receiver: NMLVDISPINFO
	arg[0]: Message selector: offsetOf: arguments: #(lvItem )

{1A7910F0: cf 1A7910CD, sp 1A79110C, bp 1A7910EC, ip 7, NMLVDISPINFO class>>itemFromNMHDR:}
	receiver: NMLVDISPINFO
	arg[0]: a ExternalAddress

{1A7910CC: cf 1A7910AD, sp 1A7910E4, bp 1A7910C8, ip 5, ListView(IconicListAbstract)>>nmGetDispInfo:}
	receiver: a ListView
	arg[0]: a ExternalAddress

{1A7910AC: cf 1A791089, sp 1A7910C0, bp 1A7910A4, ip 27, ListView>>nmNotify:}
	receiver: a ListView
	arg[0]: a ExternalAddress
	stack temp[0]: -150

### DIFF
--- a/Core/Object Arts/Dolphin/Lagoon/BasicImageStripper.cls
+++ b/Core/Object Arts/Dolphin/Lagoon/BasicImageStripper.cls
@@ -5,7 +5,7 @@ Model subclass: #BasicImageStripper
 	classVariableNames: ''
 	poolDictionaries: ''
 	classInstanceVariableNames: ''!
-BasicImageStripper guid: (GUID fromString: '{51CE4971-61E8-431F-9823-437CF7293E8A}')!
+BasicImageStripper guid: (GUID fromString: '{51ce4971-61e8-431f-9823-437cf7293e8a}')!
 BasicImageStripper comment: 'Base image stripping methods. These can safely be encrypted during release.'!
 !BasicImageStripper categoriesForClass!MVP-Models! !
 !BasicImageStripper methodsFor!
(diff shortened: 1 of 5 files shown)


blairmcg:
@jgfoster, I'm not sure I understand why this would work. The structure templates are normally removed in a deployed app, so offsetOf: will still fail, although not in the same way. It will just return -1. Actually this could well explain why the other problem with empty lists occurs in deployed apps. NMLVDISPINFO class>>itemFromNMHDR: is missing a ##, i.e. it should be:
```
itemFromNMHDR: anExternalAddress
	^LVITEM fromAddress: anExternalAddress yourAddress + ##(self offsetOf: #lvItem)
```
Can you please try adding the ## and seeing if both problems go away?

blairmcg:
As mentioned in my comment on PR #342, I've repro'd the problem with the Video Library sample and the underlying problem is the missing ## in front of the (self offsetOf: #lvitem) in the NMLVDISPINFO class>>itemFromNMHDR: method. I've pushed commit f6122bb to fix both issues. However the `ExternalStructure class>>#offsetOf:` should not be stripped if unless compiling external structures (the default), so I'll modify your PR if I can push changes to it.

PR #344, opened 2017-03-16
Update GUID and file-out version for all files to avoid “noise” 

jgfoster:
This commit has no other changes.


Core/Contributions/ITC Gorisek/ODB8ByteFloat.cls
```diff
@@ -1,11 +1,11 @@
-"Filed out from Dolphin Smalltalk X6.2"!
+"Filed out from Dolphin Smalltalk 7"!
 
 ODBDeserializers subclass: #ODB8ByteFloat
 	instanceVariableNames: ''
 	classVariableNames: ''
 	poolDictionaries: ''
 	classInstanceVariableNames: ''!
-ODB8ByteFloat guid: (GUID fromString: '{8162EE0A-0FB8-11D4-B091-000000000000}')!
+ODB8ByteFloat guid: (GUID fromString: '{8162ee0a-0fb8-11d4-b091-000000000000}')!
```
blairmcg: I just happened to notice while scrolling through that quite a lot of the classes in this package have the same guid, which obviously they shouldn't, although for most purposes it doesn't matter.

PR #350, opened 2017-03-20
Fix walkback during image stripping due to new instvar in InternalIcon

### LINKED ISSUE #349: InternalIcon instVar 'tileColor' breaks runtime deployment
Commit 8cd440380898cc2a24fc1005c8ab60b868a8f154 (#346) added the instance variable 'tileColor' to the class InternalIcon. Unfortunately, that causes a walkback during image stripping. The problem can be observed with the VideoLibrary sample.

I suspect that the changed schema causes problems with reading from a binary format.

jgfoster:
Handle change in binary filing class with new instance variable.
The actual code is based on that in the Model class.
Fix #349.

### DIFF
--- a/Core/Object Arts/Dolphin/IDE/Base/InternalIcon.cls
+++ b/Core/Object Arts/Dolphin/IDE/Base/InternalIcon.cls
@@ -196,7 +196,7 @@ stbConvert: anArray fromVersion: anInteger
 stbConvertFrom: anSTBClassFormat 
 	"Private - Convert from previous version resource. 
 	Version Changes:
-		1: Added 'tileColor' instance variable to Model."
+		1: Added 'tileColor' instance variable to InternalIcon (#346, 349)."
 
 	^
 	[:data | 
--- a/Core/Object Arts/Dolphin/IDE/Base/InternalIcon.cls
+++ b/Core/Object Arts/Dolphin/IDE/Base/InternalIcon.cls
@@ -182,7 +182,46 @@ imageManager
 	^IconImageManager current!
 
 new
-	^self withExtent: self defaultExtent! !
+	^self withExtent: self defaultExtent!
+
+stbConvert: anArray fromVersion: anInteger 
+	"Private - Convert from earlier version view by updating and answering the array of instance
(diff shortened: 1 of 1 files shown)


blairmcg:
Before I merge this I'll try creating a regression test for this. Although it isn't practical to do an actual deployment as part of the test run, I think I can see how to catch this class of error (not being able to deserialize view resources)

Core/Object Arts/Dolphin/IDE/Base/InternalIcon.cls
```diff
@@ -182,12 +182,55 @@ imageManager
...
+stbConvert: anArray fromVersion: anInteger 
+	"Private - Convert from earlier version view by updating and answering the array of instance
+	variables, instVarArray. "
+
+	| instVars |
+	instVars := anArray.
+	anInteger < 1 ifTrue: [instVars := self stbConvertFromVersion0: instVars].
+	^instVars!
+
+stbConvertFrom: anSTBClassFormat
```
blairmcg: This will work, but Model has a fairly complex upgrade requirements. For simple cases like this, the pattern used in ImageManager will be adequate:
```
stbConvertFrom: anSTBClassFormat 
	"Version 1 adds backcolor and maskcolor instance variables."

	^
	[:data | 
	| newInstance |
		newInstance := self basicNew.
		1 to: data size do: [:i | newInstance instVarAt: i put: (data at: i)].
		newInstance]
```
This form will always work if adding inst vars at the end that can be left nil.
I'll merge it anyway to see if the test passes.

jgfoster: I have a question about the object upgrade process. If I have a subclass of Model, can I have my own version numbers and stbConvertFrom: methods? It seems that the answer would be "no", but that makes subclassing from Model somewhat less useful. Right?

blairmcg: This is a weakness in the original STB design. There are ways around it, however, e.g. see ListTreeView.

PR #356, opened 2017-05-12
Add tests of ListView>>newSelectionsFromEvent:

daniels220:
Finally got around to writing these. I also fleshed out the comments in the method itself. I started trying to write a "spec" at the top as you suggested, but then as I went to comment the individual statements I found I was repeating myself. Hopefully those comments make the behavior sufficiently clear.

This seems like a good time to mention something I found in the process of writing the method initially. In a comment at the bottom, I note that the behavior of ListViews in Dolphin differs from that in e.g. Windows Explorer in one particular case. When using a Ctrl+Shift+click to _de_select a range, in Dolphin the item clicked is left selected (whether or not it previously was). I would intuitively expect it to be deselected along with everything else, and indeed this is how other programs behave. Do you have any idea why Dolphin differs in this regard?

### DIFF
--- a/Core/Object Arts/Dolphin/MVP/MultiSelectListViewTest.cls
+++ b/Core/Object Arts/Dolphin/MVP/MultiSelectListViewTest.cls
@@ -17,6 +17,116 @@ initializePresenter
 sortSelections
 	^#(49 50)!
 
+testNewSelectionsClickOutsideListWithModifiers
+	| event |
+	self setUpForSelectionTesting.
+	presenter selectionsByIndex: #(1).
+	(OrderedCollection new)
+		add: #(#control);
+		add: #(#shift);
+		add: #(#control #shift);
+		do: 
+				[:modifiers |
+				event := self mouseDownEventOnItem: 0 buttons: (modifiers copyWith: #left).
+				self assert: (presenter newSelectionsFromEvent: event) = #(1)].!
+
+testNewSelectionsCtrlClickOnSelectedItem
+	| event |
+	self setUpForSelectionTesting.
+	presenter selectionsByIndex: #(1 2).
+	event := self mouseDownEventOnItem: 1 buttons: #(#left #control).
+	self assert: (presenter newSelectionsFromEvent: event) = #(2).
+	presenter selectionsByIndex: #(1).
+	self assert: (presenter newSelectionsFromEvent: event) = #().!
+
+testNewSelectionsCtrlClickOnUnselectedItem
+	| event |
+	self setUpForSelectionTesting.
+	event := self mouseDownEventOnItem: 1 buttons: #(#left #control).
+	self assert: (presenter newSelectionsFromEvent: event) = #(1).
+	presenter selectionsByIndex: #(2).
+	self assert: (presenter newSelectionsFromEvent: event) = #(1 2).!
+
+testNewSelectionsCtrlShiftClickAdditive
+	| event |
+	self setUpForSelectionTesting.
+	presenter
+		selectionsByIndex: #(1 3);
+		selectionMarkIndex: 3.
+	event := self mouseDownEventOnItem: 7 buttons: #(#left #control #shift).
+	self assert: (presenter newSelectionsFromEvent: event) = #(1 3 4 5 6 7).
+	presenter selectionsByIndex: #(1 3 5 7).
+	self assert: (presenter newSelectionsFromEvent: event) = #(1 3 4 5 6 7).!
+
+testNewSelectionsCtrlShiftClickSubtractive
+	| event |
+	self setUpForSelectionTesting.
+	presenter
+		selectionsByIndex: #(1 2 3 4 5 7);
+		selectionMarkIndex: 6.
+	event := self mouseDownEventOnItem: 2 buttons: #(#left #control #shift).
+	"In Windows Explorer and other applications that use ListView or a similar control,
+	(2) would not be selected at this point. Why is it in Dolphin?
+	(See comment at the end of #newSelectionsFromEvent:.)"
+	self assert: (presenter newSelectionsFromEvent: event) = #(1 2 7).
+	presenter selectionsByIndex: #(1 3 5 7).
+	self assert: (presenter newSelectionsFromEvent: event) = #(1 2 7).!
+
+testNewSelectionsInitialShiftClick
+	| event |
+	self setUpForSelectionTesting.
+	event := self mouseDownEventOnItem: 1 buttons: #(#left #shift).
+	self assert: (presenter newSelectionsFromEvent: event) = #(1).
+	event := self mouseDownEventOnItem: 3 buttons: #(#left #shift).
+	self assert: (presenter newSelectionsFromEvent: event) = #(3).!
+
+testNewSelectionsNonLeftClicksWithModifiers
+	self setUpForSelectionTesting.
+	presenter selectionsByIndex: #(1 3 5).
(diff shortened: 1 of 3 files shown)


Core/Object Arts/Dolphin/MVP/MultiSelectListViewTest.cls
```diff
@@ -17,6 +17,116 @@ initializePresenter
...
+	self assert: (presenter newSelectionsFromEvent: event) = #(2).
+	presenter selectionsByIndex: #(1).
+	self assert: (presenter newSelectionsFromEvent: event) = #().!
+
+testNewSelectionsCtrlClickOnUnselectedItem
+	| event |
+	self setUpForSelectionTesting.
+	event := self mouseDownEventOnItem: 1 buttons: #(#left #control).
+	self assert: (presenter newSelectionsFromEvent: event) = #(1).
+	presenter selectionsByIndex: #(2).
```
blairmcg: Test of Ctrl+click to add a new selection item would be slightly better if the clicked item was 3+ so that there is a gap in the new selection range. This test as is would might pass even if the implementation mixed up ctrl and shift handling.

Core/Object Arts/Dolphin/MVP/MultiSelectListViewTest.cls
```diff
@@ -17,6 +17,116 @@ initializePresenter
...
+	self assert: (presenter newSelectionsFromEvent: event) = #(1 2).!
+
+testNewSelectionsCtrlShiftClickAdditive
+	| event |
+	self setUpForSelectionTesting.
+	presenter
+		selectionsByIndex: #(1 3);
+		selectionMarkIndex: 3.
+	event := self mouseDownEventOnItem: 7 buttons: #(#left #control #shift).
+	self assert: (presenter newSelectionsFromEvent: event) = #(1 3 4 5 6 7).
```
blairmcg: This is a good test as this behaviour only applies if Ctrl is down. Shift only removes the selection of 1. Do you have a test of the shift only behaviour of the result being selection between the mark and clicked item? Maybe I just haven't spotted that yet.

Core/Object Arts/Dolphin/MVP/MultiSelectListViewTest.cls
```diff
@@ -17,6 +17,116 @@ initializePresenter
...
+	presenter selectionsByIndex: #(1 3 5 7).
```
blairmcg: It's not clear to me what the selection mark would be at this point, although judging from the expected result it must still be `3`? This implies the behaviour of the control is to preserver the previous selection mark. Might be good to assert that the selection mark is still what you expect after changing the selection, if only for understanding purposes.

Core/Object Arts/Dolphin/MVP/MultiSelectListViewTest.cls
```diff
@@ -17,6 +17,116 @@ initializePresenter
...
+	self assert: (presenter newSelectionsFromEvent: event) = #(1 3 4 5 6 7).!
+
+testNewSelectionsCtrlShiftClickSubtractive
+	| event |
+	self setUpForSelectionTesting.
+	presenter
+		selectionsByIndex: #(1 2 3 4 5 7);
+		selectionMarkIndex: 6.
+	event := self mouseDownEventOnItem: 2 buttons: #(#left #control #shift).
+	"In Windows Explorer and other applications that use ListView or a similar control,
```
blairmcg: As mentioned earlier, probably a separate bug that we ought to fix. That will be easier to do without breaking other selection behaviour now you have put in place these tests.

Core/Object Arts/Dolphin/MVP/MultiSelectListViewTest.cls
blairmcg: Yes, `testNewSelectionsShiftClickWithSelectionMark` covers that well.

Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
```diff
@@ -1474,41 +1474,51 @@ moreParamsAt: n put: anObject
...
+	otherwise it would be very frustrating to open the context menu with multiple items selected.
+	The control also chooses to leave the selection alone if modifiers are held during a right/middle click."
+	(event button ~= #left and: [anyKeysDown or: [newSelections includes: itemClicked]])
+		ifTrue: [^newSelections].
+	"Having taken care of the above edge cases, if no keys are down, the behavior is equivalent to the single-select case."
 	anyKeysDown ifFalse: [^itemClicked ifNil: [#()] ifNotNil: [:item | Array with: item]].
-	"Only Ctrl"
+	"Only Ctrl is down, so we flip the selection state of the clicked item."
 	event isShiftDown ifFalse: 
 			[newSelections := newSelections asSortedCollection.
```
blairmcg: Not code you are changing here, but is it ever the case that the `selectionsByIndex` returns an unordered list? I don't think so as `ListView>>getMultipleSelections` says the selections will be in ascending order. Sorting an already ordered list is a pathologically slow case for quicksort, although probably this wouldn't be an issue in practice as Dolphin uses a modified quicksort by default, and the number of selected items is not likely to be very large. A good reason to use a `SortedCollection` would be to take advantage of an optimised `remove:ifAbsent:`, e.g. using a binary search, but currently the implementation isn't optimised and will perform as for `OrderedCollection`.

daniels220: No, I don't think `selectionsByIndex` should never return an unsorted list. The purpose of making a `SortedCollection` is so that, in the case where we are _adding_ a selection, the value we ultimately return is in sorted order. I admit I didn't think of the pathological behavior of quicksort in this case, but I agree that the combination of Dolphin's mitigation and the (usually) small size of the collection make it not that important. 

I have often wondered about implementing an optimized `remove:ifAbsent:` on SortedCollection...something to stick in the back of my/our mind(s), I guess.

Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
```diff
@@ -1474,41 +1474,51 @@ moreParamsAt: n put: anObject
...
+	and everything in between it and the selection mark."
 	newSelections := newSelections asSet.
 	markIsSelected := newSelections includes: markIndex.
 	markRange do: (markIsSelected
 				ifTrue: [[:i | newSelections add: i]]
 				ifFalse: [[:i | newSelections remove: i ifAbsent: []]]).
-	"For some reason, the item clicked is always left selected,
-	even if we are deselecting the range (markIsSelected == false).
-	This is not true in e.g. Windows Explorer--why?"
+	"For some reason, the item clicked is always left selected, even if we are deselecting the rest
```
daniels220: Hmm...any idea where I might look to find the cause? At a glance, I'm not seeing anywhere that interacts with the selection while it is changing other than `onButtonPressed:` itself.

Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
```diff
@@ -1912,6 +1922,15 @@ selectionMarkIndex
 
 	^(self sendMessage: LVM_GETSELECTIONMARK) + 1.!
 
+selectionMarkIndex: anInteger
```
daniels220: If so, I would think the caret (the item with the focus ring) should also be preserved, and it is not. I'll make another PR fixing this for both ListView and ListBox.

PR #357, opened 2017-05-14
Restore the anchor/selectionMark and caret when recreating a ListBox/ListView

daniels220:
Add `anchorIndex[:]` accessors to ListBox (this is its name for what ListView calls `selectionMarkIndex`). Add `caretIndex:` setter to ListView. Modify `state` to include restoring those if needed.

Not sure if I got `ListView>>caretIndex:` quite right...what do you think? It does seem to work in my messing around with it.

### DIFF
--- a/Core/Contributions/Solutions Software/EditableListView.cls
+++ b/Core/Contributions/Solutions Software/EditableListView.cls
@@ -469,6 +469,23 @@ itemAndColumnAt: aPoint
 		ifTrue: [itemIndex@columnIndex]
 		ifFalse: [nil]!
 
+itemIndex: anInteger hasFocus: aBoolean
+	| anLvItem |
+	#deprecated.
+	"Only one item may have focus at a time--use #caretIndex: to set it."
+	anLvItem := LVITEM new.
+	anLvItem stateMask: LVIS_FOCUSED.
+	anLvItem dwState: aBoolean.
+	self lvmSetItem: anInteger - 1 state: anLvItem.!
+
+itemIndexWithFocus
+	#deprecated.
+	^self caretIndex.!
+
+itemIndexWithFocus: anInteger
+	#deprecated.
+	self caretIndex: anInteger.!
+
 lastEditableColumn
 
 	^self allColumns reverse detect: [ :each | each isEditable]!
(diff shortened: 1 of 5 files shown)


Core/Object Arts/Dolphin/MVP/Presenters/List/ListBox.cls
```diff
@@ -443,13 +453,29 @@ showDropHighlight: anItem
...
+	messages := super state.
+	"Add the horizontal extent aspect since it is not otherwise cached in an instance variable."
+	messages add: (MessageSend
+				receiver: self
+				selector: #horizontalExtent:
+				argument: self horizontalExtent).
+	"The caret and anchor only need to be restored separately in multi-select mode--
+	in single-select mode they are always equal to the selected item."
+	self isMultiSelect ifTrue: 
+			[self anchorIndex = 0 ifFalse:
```
blairmcg: Given that `anchorIndex` and `caretIndex` are computed by sending a message to the control (which will call back into Dolphin, before being passed back to the control's own window procedure), it would be worthwhile storing the returned values into temps to avoid having to compute each twice.

Core/Contributions/Solutions Software/EditableListView.cls
```diff
@@ -475,28 +469,6 @@ itemAndColumnAt: aPoint
 		ifTrue: [itemIndex@columnIndex]
 		ifFalse: [nil]!
 
-itemIndex: anInteger hasFocus: aBoolean
```
blairmcg: What was the motivation for removing these completely? I'm not familiar with this control, but the method categorisation suggests they were considered part of the public interface.

Core/Object Arts/Dolphin/MVP/Presenters/List/ListBox.cls
```diff
@@ -17,6 +17,16 @@ Class Variables:
 !ListBox categoriesForClass!MVP-Views! !
 !ListBox methodsFor!
 
+anchorIndex
```
blairmcg: The term "anchor" seems preferable to "selection mark". Perhaps we should just settle on this terminology and use the same selector names for ListView as well, that is if the anchor and selection mark behave in exactly the same way (I assume they do?). What do you think?

daniels220: I remember coming up with the term "anchor" to describe the concept, early on when I was investigating the behavior of the control, before even looking at the documentation and finding the term "selection mark". So yes, I find it preferable also.

I was also vaguely wondering, as I made the changes, whether it would be worth standardizing, but for whatever reason didn't do so. Since you independently thought of the same idea, I gave it a shot, and it turns out to be totally worth it—it becomes possible to push up the `#state` changes to ListControlView, which saves a lot of code. (I do believe the ListBox "anchor" and ListView "selection mark" behave the same, yes.)

Core/Object Arts/Dolphin/MVP/Presenters/List/ListBox.cls
daniels220: Incorporated into the common implementation on ListControlView.

Core/Contributions/Solutions Software/EditableListView.cls
daniels220: `itemIndexWithFocus[:]` was an (inefficient) duplicate of `caretIndex[:]`, and `itemIndex:hasFocus:` misunderstands the nature of the LVIS_FOCUSED style—though it appears to be a per-item style, the control enforces that it may only be applied to one item at a time.

Fair point, though, that these may be public API—I added them back with `#deprecated` tagging, aliasing `itemIndexWithFocus[:]` to `caretIndex[:]` and explaining why `itemIndex:hasFocus:` does not allow any behavior that simply setting the `caretIndex` does not.

PR #360, opened 2017-05-21
Load and save all loadable packages

jgfoster:
This should update the GUID, date, and fileout version so that subsequent saves on a package will not have extra "noise". There are a few (3) packages that exist but don't load and a number of classes that are in the file system but not saved through this process. I'll investigate them next.


Core/Object Arts/Dolphin/Utilities/IPHelp/IP Help API.pax
```diff
@@ -59,6 +59,7 @@ package globalAliases: (Set new
 
 package setPrerequisites: (IdentitySet new
 	add: '..\..\Base\Dolphin';
+	add: '..\..\Base\Deprecated\Dolphin Base (Deprecated)';
```
blairmcg: I'll raise an issue for this. Obviously no packages should be depending on the deprecated ones.

Core/Object Arts/Dolphin/System/Announcements/Announcement.cls
```diff
@@ -1,29 +1,29 @@
...
+Object subclass: #Announcement
+	instanceVariableNames: ''
+	classVariableNames: ''
+	poolDictionaries: ''
+	classInstanceVariableNames: ''!
+Announcement guid: (GUID fromString: '{12525b20-16ef-4684-b01e-956b227a429d}')!
+Announcement comment: 'This class is the superclass for events that someone might want to announce, such as a button click or an attribute change. Typically you create subclasses for your own events you want to announce.

 
'!
```
blairmcg: There are bad line endings in all these Announcement related classes. I think it must have been filed in from Squeak/Pharo before the support was added for correctly reading mixed line endings in file-ins. I'll raise an issue for this.

PR #367, opened 2017-05-22
Remove events for model when it is replaced

jgfoster:
It seems to me that we should be able to replace the model for a presenter and stop getting notifications for changes to a prior model.
ValuePresenter is the main culprit, but I looked at other implementers of #’model:’ that add events.
See ViewTest>>testReplaceView for an example.

I accept that I may have a misunderstanding of how the MVP framework works, but when I have a Presenter on a more complex object (subclassing from Model), I don't want changes to other objects to be falsely reported as being on the current object.


blairmcg:
I basically agree, but I think this be closed to pattern used in View, i.e. there should be `Presenter>>disconnectFromModel` with the same implementation as the view version. As to whether `Presenter>>model:` should send this by default, I'd be concerned about the potential performance impact - `EventsCollection>>removeEventsTriggeredFor:` is a serial search within a process synchronisation lock, so quite expensive.
Since the `Presenter>>createSchematicWiring` does nothing by default, I think not doing the disconnect by default is not unreasonable. Specific subclasses that do observe the view (I assume you caught all the cases) can then do the `disconnectFromModel` in the same places you've modified. 
It would be preferable from a framework ease of use perspective to put the `disconnectFromModel` call into `Presenter>>model:`, but one would need to measure the perf impact. Actually the same may apply to modifying ValuePresenter>>model:, so I'd want to check before/after timings for opening a variety of different views, especially the more complex browsers, or dialogs with a lot of fields.

jgfoster:
Thanks for the feedback. I'll investigate the alternatives.

PR #368, opened 2017-05-22
Remove packages and classes that can’t be loaded

jgfoster:
- EnumRect Sample references two classes that are already loaded.
- Dolphin Socket Tests references a few classes that don’t exist.
- Dolphin Application Automation has a prerequisite package that doesn’t exist.
I think this was missed from yesterday's submissions.


blairmcg:
I think all of these, and especially the sockets tests, should potentially be restored rather than deleted.

PR #374, opened 2017-05-23
Ability to get user session idle time

jgfoster:
How long has the user been idle? This allows us to deal with user inactivity on the system, not just in this application.


Core/Object Arts/Dolphin/MVP/Base/Dolphin MVP Base.pax
```diff
@@ -1275,6 +1283,23 @@ icon
...
+
+userSessionIdleMs
+"
+	(Delay forSeconds: 1) wait. SessionManager inputState userSessionIdleMs.
+"
+	| currentTime flag lastInputInfo lastInputTime |
+	lastInputInfo := LASTINPUTINFO new.
+	flag := UserLibrary default getLastInputInfo: lastInputInfo.
+	flag ifFalse: [self error: 'Error obtaining last input time'].
+	currentTime := self millisecondClockValue.
```
blairmcg: From MSDN it seems that `GetLastInputInfo` returns a millisecond clock value from `GetTickCount`. Dolphin's millisecond clock primitive is _not_ based on this timer. It's not based on the WinMM timer any more either (the comment `InputState>>millisecondClockValue` is wrong). In fact it uses the high precision CPU clock via the `QueryPerformanceCounter` API, converted to a millisecond value. This may or may not produce a value that has the same baseline as `GetTickCount`; testing on my machine just now, the QPC derived clock reported by the primitive seems to be about 1.2 seconds ahead of the millisecond clock reported by a call to `GetTickCount`. 
In conclusion I think you need to add `KernelLibrary>>getTickCount` and then use that instead to initialize `currentTime`.
```
!KernelLibrary methodsFor!

getTickCount
	<stdcall: dword GetTickCount>
	^self invalidCall! !
!KernelLibrary categoriesFor: #getTickCount!public!win32 functions-date and time! !
```

jgfoster: I've made the suggested changes and updated the comment!

Core/Object Arts/Dolphin/Base/KernelLibrary.cls
```diff
@@ -1686,6 +1710,7 @@ zeroMemory: pvDest length: cBytes
 !KernelLibrary categoriesFor: #getTempFileName:lpPrefixString:uUnique:lpTempFileName:!public!win32 functions-file! !
 !KernelLibrary categoriesFor: #getTempPath:lpBuffer:!public!win32 functions-file! !
 !KernelLibrary categoriesFor: #getThreadTimes:lpCreationTime:lpExitTime:lpKernelTime:lpUserTime:!public!win32 functions-process and thread! !
+!KernelLibrary categoriesFor: #getTickCount!public!win32 functions-natural language support! !
```
blairmcg: Wrong category, but we can fix that later

jgfoster: Indeed! I've selected more reasonable categories and committed a change. Now to see if it shows up as a new pull request...

PR #376, opened 2017-05-27
Various improvements to autocomplete and the search box

daniels220:


Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
```diff
@@ -154,47 +151,47 @@ rangeToCompleteAt: anInteger
...
 searchItemsStartingWith: aString maxItems: anInteger
-	| result classes filter |
-	aString first isLowerCase
-		ifTrue: 
-			[| matchingSymbols |
-			matchingSymbols := self selectorsStartingWith: aString maxItems: anInteger.
-			^matchingSymbols].
-	filter := (aString includes: $>)
+	| result classes classFilter addBlock |
+	aString first isUppercase ifFalse: [^self selectorsStartingWith: aString maxItems: anInteger].
```
blairmcg: Is the motivation for this change to switch from looking for any lowercase letter to any non-uppercase character (which would therefore include underscore)? If so it might be worth noting that in a comment, as it is a bit obscure otherwise.

Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
```diff
@@ -154,47 +151,47 @@ rangeToCompleteAt: anInteger
 				ifFalse: [anInteger to: view caretPosition - 1]]!
 
 searchItemsStartingWith: aString maxItems: anInteger
```
blairmcg: Can you note here (not in the code, I mean as a self-review comment) what the effect of the changes to this method are please. It's not easy to tell from inspection.

Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
```diff
@@ -154,47 +151,47 @@ rangeToCompleteAt: anInteger
...
 	searchText := self searchText.
-	searchText first isLowerCase ifTrue: [^Symbol findInterned: searchText].
-	evaluateText := searchText copyReplaceAll: '>>' with: '>>#'.
-	Compiler evaluate: evaluateText] 
-			on: Exception
-			do: [:x | Package manager packages detect: [:each | each name = self searchText] ifNone: []]!
+	searchText isEmpty ifTrue: [^nil].
+	(searchText first isUppercase not and: [Symbol isLiteralSymbol: searchText])
+		ifTrue: [^Symbol findInterned: searchText].
+	(Package manager packageNamed: searchText ifNone: []) ifNotNil: [:a | ^a].
```
blairmcg: Here you've effectively changed the search order so that packages are matched before any same-named expression. What was the reason to do that? (not saying it is wrong to change it, just wondering)

Core/Object Arts/Dolphin/IDE/Base/SmalltalkWorkspace.cls
```diff
@@ -902,46 +909,20 @@ messagesForToken: anAssociation startingWith: aString maxItems: anInteger
...
-							eachClass selectors do: 
-									[:eachSelector |
-									(eachSelector beginsWith: aString ignoreCase: ignoreCase)
-										ifTrue: 
-											[messages add: eachSelector.
-											messages size > anInteger ifTrue: [^#()]]]]]).
-			messages])
-			collect: [:each | ScintillaListItem text: each icon: icon]!
+	class := self classForToken: anAssociation.
+	class == UndefinedObject ifTrue: [class := nil].
```
blairmcg: Why would you want to treat this as a special case? If I type nil in a workspace, I'd expect to be offered only messages understood by UndefinedObject in the auto-completion list.

Core/Object Arts/Dolphin/IDE/Base/SmalltalkWorkspace.cls
```diff
@@ -902,46 +909,20 @@ messagesForToken: anAssociation startingWith: aString maxItems: anInteger
...
+	selectors := Set new.
+	Smalltalk developmentSystem allSelectors do:
```
blairmcg: This is simpler, but on inspection it doesn't appear to change the behaviour. It does affect the performance characteristics in some extreme cases, however, especially when the prefix is empty (e.g. when you press space after a potential receiver token). Two cases are:
1. When the receiver class is not known (the normal case), if there is no prefix then the old code will quickly early out and return no choices because there are too many (line 913).
1. When the receiver class is known (e.g. for a literal, a less common case), the old code would early out without much work other than collecting the understood selectors if the number of potential matches exceeded the maximum allowed.

It also looks to me like the old code will perform far less work for short prefixes for typical known receiver classes.
What is the motivation to change this method anyway?

Core/Object Arts/Dolphin/IDE/Base/SmalltalkWorkspace.cls
daniels220: The motivation for this was the case where a _variable_ currently has the value `nil`, e.g. immediately after restarting a frame in the Debugger, or perhaps after setting a pool variable to nil in a workspace. In this case the nil-ness is likely temporary, and I've found it more useful to be offered everything rather than almost nothing. Good point about the literal case, though—I could make the special case narrower, so it only applies if the receiver is a variable, not the literal `nil`.

Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
daniels220: Generally, an intuitive sense that since there are fewer packages than classes in the system, a search that matches one exactly should return it rather than a class of the same name. The very _specific_ motivation, and the reason I initially made the change, was the name conflict between the _package_ `RBParser` and the _global_ of the same name. (The rest of the implementation makes a "class > package > global" implementation difficult—"class" and "global" are both covered by the `Smalltalk at:` line, and teasing the two apart seems more trouble than it's worth.)

Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
daniels220: Yes, precisely. (Other punctuation is also included, which allows searching for binary selectors.) I'll add a comment.

Core/Object Arts/Dolphin/IDE/Base/SmalltalkWorkspace.cls
daniels220: There is a (subtle but important) change in behavior, which in turn is the motivation for the change. The new code is willing to consider selectors that the class does not in fact "implement" in the usual sense, so long as it claims that it `#canUnderstand:` them (presumably via a `#doesNotUnderstand:` implementation). I realize this does depend on those selectors being present in the `allSelectors` collection in the first place—I've added a commit changing `#allSelectors` to use `#understoodSelectors`, giving such classes a chance to ensure this is the case.

Regarding the performance cases—all true. I guess the bottom line for me is that:
1. The worst-case is no worse than before—completing against a receiver of unknown type already examined everything, and is as you say the common case anyway.
2. Performance of code triggered only by user interaction (not potentially run in a tight loop) is just so much less of an issue than it was 10 years ago that I think the simplicity and ease of understanding is worth the performance cost, such as it is.

Core/Object Arts/Dolphin/IDE/Base/SmalltalkSearchPresenter.cls
daniels220: Sure:

* As you noticed, `isLowercase ifTrue:` -> `isUppercase ifFalse:`—treat searches starting with punctuation as referring to selectors.
* Add packages to the list ahead of classes, reflecting their higher priority when the search is actually performed.
* Scan `Smalltalk` for source and alias globals and add them to the list also.
* Actually respect the `maxItems:` parameter for non-selector searches.
* Revert to using string equality to compare class names, as _metaclass_ names are strings, not symbols.

Core/Object Arts/Dolphin/IDE/Base/SmalltalkWorkspace.cls
```diff
@@ -902,46 +909,22 @@ messagesForToken: anAssociation startingWith: aString maxItems: anInteger
...
-							[:eachClass |
-							eachClass selectors do: 
-									[:eachSelector |
-									(eachSelector beginsWith: aString ignoreCase: ignoreCase)
-										ifTrue: 
-											[messages add: eachSelector.
-											messages size > anInteger ifTrue: [^#()]]]]]).
-			messages])
-			collect: [:each | ScintillaListItem text: each icon: icon]!
+	class := self classForToken: anAssociation.
```
blairmcg: I can't seem to reply to your final comment so sorry that this is a new thread. 
In general I agree about the perf argument, except that it isn't just about speed/responsiveness, its also whether the battery on one's laptop is being flattened for no productive reason. The particular cases I would care about (that I perhaps didn't adequately explain) is where the list is gong to be too long anyway and will not be displayed to the user, especially where the prefix string is empty. I think it is still worth special casing that (as in the old code lines 912/3) so that when the user presses the space bar after an identifier of unknown type it doesn't enumerate the `allSelectors` collection at all.

blairmcg: Ignore that. You are right, it isn't worth worrying about. As the maximum number of items is relatively small, the amount of computation is small and the loop takes <1mS to run for an empty prefix anyway.

daniels220: I was thinking that was likely to be the case, but was going to test it myself (always test! :) I've learned that the hard way a number of times...) before getting back to you—thanks for checking it out.

PR #383, opened 2017-06-24
Update EditableListView to latest version

rko281:
Refactoring/rewrite, including use of new base ListView methods (caretIndex: etc.). Also improves list behaviour to be more standard (display, multi select, keyboard tabbing).


rko281:
Appears to be missing a class file - will retract and review.

PR #384, opened 2017-06-24
Update EditableListView to latest version

rko281:
Refactoring/rewrite, including use of new base ListView methods (caretIndex: etc.). Also improves list behaviour to be more standard (display, multi select, keyboard tabbing).

[Attempt 2; added missing EmbeddedCheckButton.cls file]

### DIFF
--- a/Core/Contributions/Solutions Software/SSW Widget Enhancements.pax
+++ b/Core/Contributions/Solutions Software/SSW Widget Enhancements.pax
@@ -1,7 +1,7 @@
 | package |
 package := Package name: 'SSW Widget Enhancements'.
 package paxVersion: 1;
-	basicComment: '© Solutions Software/John Aspinall 2016
+	basicComment: '© Solutions Software/John Aspinall 2017
 Provided under the same licence terms as Dolphin Smalltalk 7'.
 
 package basicPackageVersion: '0.001'.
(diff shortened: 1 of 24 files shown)


Core/Contributions/Solutions Software/SSW Widget Enhancements.pax
```diff
@@ -106,6 +122,15 @@ TextEdit subclass: #FormattedTextEdit
 
 !Canvas methodsFor!
 
+clipTo: aRectangle during: aBlock
+
+	| oldClipRegion |
+
+	oldClipRegion := Region empty.
+	GDILibrary default getClipRgn: self asParameter hrgn: oldClipRegion asParameter.
```
blairmcg: I had to think hard to convince myself that this doesn't leak a region, but I think it is OK. You might want to explicitly free oldClipRegion at the end of your ensure block, however, in order to avoid generating finalizable garbage.

Core/Contributions/Solutions Software/SSW Widget Enhancements.pax
```diff
@@ -303,6 +346,41 @@ resendTo: aView
...
+	style = BS_CENTER ifTrue: [^#center].
+	style = BS_RIGHT ifTrue: [^#right].
+
+	^#center!
+
+alignment: aSymbol
+
+	| mask |
+
+	self baseStyleMask: BS_CENTER set: false.
```
blairmcg: This might end up creating the control twice as it is changing the base style twice. It looks like this could use View>>baseStyle:maskedBy: (e.g. as in TextEdit>>setAlignment:)? I realise it is private, but it shouldn't be.

PR #390, opened 2017-07-03
Fix RefactoringSmalltalkSystem>>renameTemporary:to:in:selector: to us…

wdargel:
…e #return: on an exception rather than no longer existing #exitWith:.


blairmcg:
Sorry it has taken me several days to get to this.

PR #392, opened 2017-07-07
Small Windows HTTP Server package improvements

blairmcg:


fxgallego:
@blairmcg using this server I was able to get 3000 req/sec in example1 removing the log to transcript, pretty impresive. And it appears to be stable without memory consumption. I think that it can be used as replacement for swazoo. Do you have any comments about your motivation to include this package?

blairmcg:
@fxgallego, I started working on it because Andy needed a high performance HTTP server. It didn't get very far because the need went away, and also because I spent more time improving the type library code generation in order to generate a better wrapper than I did working on the package.
I tend to think of external component wrappers having 3 levels:

1. Basic ExternalLibrary library wrapper for the API, and ExternalStructures for any structures it uses (in this case, a lot). This is a low-level layer, and hard to use. I generally try to auto-generate these from a type library, building one of those if necessary by creating IDL from the C header files. Doing a good job of this requires adding quite a lot of extra metadata attributes to the information in the .h files. The advantage of doing this is that regenerating the wrapper to reflect changes is easy, and it avoids mistakes in sizing, choice of external types, etc, that often result in memory corruption and crashes when trying to use the API.
1. Hand-written extensions and refactorings of the first level wrapper to make it more usable from Smalltalk, e.g. marshalling between standard Smalltalk types and byte structures etc. This layer should be easier and more natural to use, but still doesn't have any higher-level abstractions, so is inconvenient and verbose.
1. Proper abstractions for the contents in the library that hide all the external interfacing details so that the consuming developer is not aware that the component isn't implemented in entirely Smalltalk, and has convenient high-level classes that are easy and natural to use.

I would say this work is somewhere between 1 and 2. I have a type library that defines the complete API (this isn't pushed to GitHub). I'd started experimenting with the generated API and adding bits of level 2 along with the examples. I didn't explore the whole API, so its likely that some parts of the generated API have not had all the necessary metadata added to the typelib to generate the best quality low-level API functions/structures. I also haven't fleshed out or written examples for all the functionality of the API/component, as it is quite extensive. I haven't even started on higher-level abstractions - e.g. it needs a proper HttpRequest class that hides most of the details.

There is still a lot of work to do to make this a replacement for Swazoo, although maybe it would be possible to replace some of the lower-level layers of Swazoon with this (I don't know how practical that would be).

fxgallego:
@blairmcg thanks for your detailed explanation.

We currently use swazoo to serve a REST API. We've patched swazoo in several places in order to rid off several bugs causing crashes and hangs. Sadly, we can't figure out the reason of this problems because they aren't easy to reproduce so we don't know in which part is the problem, could it be in process, error handling, sockets, semaphores.

We'll try to implement our API using this server. Maybe we can borrow some ideas from Swazoo for the 3rd level layer or directly replace the swazoo low level server loop. I let you know if we need some extra functionality that's supported by the api and currently not implemented in Dolphin.

fxgallego:
BTW, I've found an easy way to reproduce the Dolphin crash with a swazoo server being stressed by the apache benchmark tool. Below the error details. This two errors where generated by two different sessions. I don´t want to report an issue for this but perhaps this helps to identify a hidden bug. Please, don't hesitate to ask me for more details.
```
********************************************************************************
************************** Dolphin Crash Dump Report ***************************

10:04:17, 10/10/2017: Dolphin7.exe caused an unhandled Win32 Exception C0000005
at 10002C02 in module 10000000 (C:\Users\scalvo\dev\Dolphin\DolphinVM7.DLL)

*----> Exception Parameters <----*
00000001	
0697FFFC	

*----> CPU Context for thread 0x1c20 <----*
EAX = 06980000 EBX = 10060390 ECX = 1078AA30
ESI = 10787810 EDI = 10787810 EIP = 10002C02
ESP = 00C0F774 EBP = 00C0F790 EFL = 00210246
CS = 0023 SS = 002B DS = 002B
ES = 002B FS = 0053 GS = 002B

*----> VM Context <----*
Process: {C17F0210:size 276365657 words, suspended frame 0363075D, priority 3, callbacks 0
last failure 0:nil, FP control 80017, thread 00000000}
Active Method: WS2_32Library>>accept:addr:addrlen:
IP: 070B68B0 (0)
SP: 03630784
BP: 03630750 (276365633)
ActiveFrame: {0363075C: cf 1078B150, sp 00000000, bp 03630750, ip 0, WS2_32Library>>accept:addr:addrlen:}
	receiver: a WS2_32Library
	arg[0]: a ExternalHandle(00000860)
	arg[1]: a SOCKADDR_IN
	arg[2]: a DWORD


New Method: WS2_32Library>>accept:addr:addrlen:
Message Selector: accept:addr:addrlen:

*----> Stack Back Trace <----*
{0363075C: cf 1078B150, sp 00000000, bp 03630750, ip 0, WS2_32Library>>accept:addr:addrlen:}
	receiver: a WS2_32Library
	arg[0]: a ExternalHandle(00000860)
	arg[1]: a SOCKADDR_IN
	arg[2]: a DWORD


***Unable to generate complete exception report (c0000005)

***** End of crash report *****


********************************************************************************
************************** Dolphin Crash Dump Report ***************************

18:44:26, 11/10/2017: Dolphin7.exe caused an unhandled Win32 Exception 80000003
at 100063B9 in module 10000000 (C:\Users\scalvo\dev\Dolphin\DolphinVM7.DLL)

*----> Exception Parameters <----*
00000000	

*----> CPU Context for thread 0x32dc <----*
EAX = 5E7F035D EBX = 0AB80210 ECX = 5E7F035C
ESI = 105ACE40 EDI = 001D40C0 EIP = 100063B9
ESP = 00C0FDB4 EBP = 00C0FDFC EFL = 00200202
CS = 0023 SS = 002B DS = 002B
ES = 002B FS = 0053 GS = 002B

*----> VM Context <----*
Process: {FF7F0210:size 398458964 words, suspended frame 5E7F0345, priority 3, callbacks 0
last failure 0:nil, FP control 80017, thread 00000000}
Active Method: ProcessorScheduler>>enableInterrupts
IP: 07274166 (6)
SP: 5E7F035C
BP: 5E7F0344 (398458942)
ActiveFrame: {5E7F0344: cf 5E7F0329, sp 5E7F035C, bp 5E7F0344, ip 7, ProcessorScheduler>>enableInterrupts}
	receiver: a ProcessorScheduler


New Method: ProcessorScheduler>>yield
Message Selector: yield

*----> Stack Back Trace <----*
{5E7F0344: cf 5E7F0329, sp 5E7F035C, bp 5E7F0344, ip 7, ProcessorScheduler>>enableInterrupts}
	receiver: a ProcessorScheduler

{5E7F0328: cf 5E7F0305, sp 5E7F033C, bp 5E7F0320, ip 2, ProcessorScheduler>>gpFault:}
	receiver: a ProcessorScheduler
	arg[0]: a ByteArray
	stack temp[0]: nil

{5E7F0304: cf 5E7F02D9, sp 5E7F0318, bp 5E7F02FC, ip 16, [] in ProcessorScheduler>>vmi:list:no:with:}
	receiver: a ProcessorScheduler

{5E7F02D8: cf 5E7F02B9, sp 5E7F02EC, bp 5E7F02D4, ip 18, BlockClosure>>ifCurtailed:}
	receiver: [] @ 0 in nil
	arg[0]: [] @ 24 in ProcessorScheduler>>vmi:list:no:with:

{5E7F02B8: cf 5E7F028D, sp 5E7F02CC, bp 5E7F02A8, ip 27, ProcessorScheduler>>vmi:list:no:with:}
	receiver: a ProcessorScheduler
	arg[0]: -553648004
	arg[1]: a LinkedList
	arg[2]: 5
	arg[3]: a ByteArray

{5E7F028C: cf 107F2570, sp 5E7F02A0, bp 5E7F028C, ip 0, Process>>primTerminate}
	receiver: a Process(HTTPConnection>>interact base 5E7F0000 in Process>>shutdown sp=5E7F0288 ip=15 list=10050010)


***Unable to generate complete exception report (c0000005)

***** End of crash report *****
```

PR #394, opened 2017-07-08
Add primality test

FunctionPoint:
Integer>>isPrime with regression test.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/Integer.cls
+++ b/Core/Object Arts/Dolphin/Base/Integer.cls
@@ -497,6 +497,23 @@ isPowerOfTwo
 
 	^self strictlyPositive and: [(self bitAnd: self - 1) = 0]!
 
+isPrime
+	"Answer true if the receiver is a prime number."
+
+	| maxDivider |
+
+	"Handle special cases."
+	self < 2 ifTrue: [ ^ false ].					"1, 0 and negative numbers are not prime."
+	( self = 2 or: [ self = 3 ] ) ifTrue: [ ^ true ].	"2 and 3 are prime."
+	self \\ 2 = 0 ifTrue: [ ^ false ].				"Other even numbers are not prime."
+
+	"Now check divisibility by all odd dividers upto the square root of self."
+	maxDivider := self sqrt asInteger + 1.
+	3 to: maxDivider by: 2 do: [ :divider |
+		self \\ divider = 0 ifTrue: [ ^ false ] ].
+
+	^ true.!
+
 isSDWORD
 	"Answer whether the receiver can be represented as a 32-bit two's complement signed integer
 	(i.e. as an SDWORD)."
@@ -768,6 +785,7 @@ truncated
 !Integer categoriesFor: #isInteger!public!testing! !
 !Integer categoriesFor: #isNull!public!testing! !
 !Integer categoriesFor: #isPowerOfTwo!public!testing! !
+!Integer categoriesFor: #isPrime!bit manipulation!public!testing! !
 !Integer categoriesFor: #isSDWORD!public!testing! !
 !Integer categoriesFor: #lcm:!factorization!public! !
 !Integer categoriesFor: #lowBit!bit manipulation!public! !
--- a/Core/Object Arts/Dolphin/Base/IntegerTest.cls
+++ b/Core/Object Arts/Dolphin/Base/IntegerTest.cls
@@ -68,6 +68,14 @@ testIsPowerOfTwo
 	powersOfTwo do: [:each | self assert: each isPowerOfTwo].
 	nonPowersOfTwo do: [:each | self deny: each isPowerOfTwo]!
 
+testIsPrime
+	| isPrimeResults |
+
+	isPrimeResults := #( 1 2 3 4 5 6 7 8 9 10 11 12 13 1000 7919 ) collect: [ :num | num isPrime ].
+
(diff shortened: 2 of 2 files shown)


Core/Object Arts/Dolphin/Base/Integer.cls
```diff
@@ -497,6 +497,23 @@ isPowerOfTwo
 
 	^self strictlyPositive and: [(self bitAnd: self - 1) = 0]!
 
+isPrime
```
blairmcg: As a general point, when adding methods to base classes please follow the standard formatting for consistency. The easiest way to do this is to auto-format with Ctrl+W, or use Reformat/Accept (Ctrl+Shift+S) when saving. Unfortunately you may find this messes up your comment alignment.

Core/Object Arts/Dolphin/Base/Integer.cls
```diff
@@ -497,6 +497,23 @@ isPowerOfTwo
...
+	"Answer true if the receiver is a prime number."
+
+	| maxDivider |
+
+	"Handle special cases."
+	self < 2 ifTrue: [ ^ false ].					"1, 0 and negative numbers are not prime."
+	( self = 2 or: [ self = 3 ] ) ifTrue: [ ^ true ].	"2 and 3 are prime."
+	self \\ 2 = 0 ifTrue: [ ^ false ].				"Other even numbers are not prime."
```
blairmcg: `self even`

Core/Object Arts/Dolphin/Base/IntegerTest.cls
```diff
@@ -68,6 +68,14 @@ testIsPowerOfTwo
 	powersOfTwo do: [:each | self assert: each isPowerOfTwo].
 	nonPowersOfTwo do: [:each | self deny: each isPowerOfTwo]!
 
+testIsPrime
+	| isPrimeResults |
+
+	isPrimeResults := #( 1 2 3 4 5 6 7 8 9 10 11 12 13 1000 7919 ) collect: [ :num | num isPrime ].
+
+	self assert: isPrimeResults =
```
blairmcg: This would provide more immediately actionable test failure information if each case were asserted individually, e.g:

```| cases |
cases := #((1 false) (2 true) (3 true) (4 false).... etc.
cases do: [:pair | self assert: pair first isPrime equals: pair second]```

File: Core/Object Arts/Dolphin/Base/Integer.cls [hunk 4]
```diff
@@ -768,6 +785,7 @@ truncated
 !Integer categoriesFor: #isInteger!public!testing! !
 !Integer categoriesFor: #isNull!public!testing! !
 !Integer categoriesFor: #isPowerOfTwo!public!testing! !
+!Integer categoriesFor: #isPrime!bit manipulation!public!testing! !
```
blairmcg: The `bit manipulation` category does not apply here.

Core/Object Arts/Dolphin/Base/Integer.cls
```diff
@@ -497,6 +497,23 @@ isPowerOfTwo
...
+
+	| maxDivider |
+
+	"Handle special cases."
+	self < 2 ifTrue: [ ^ false ].					"1, 0 and negative numbers are not prime."
+	( self = 2 or: [ self = 3 ] ) ifTrue: [ ^ true ].	"2 and 3 are prime."
+	self \\ 2 = 0 ifTrue: [ ^ false ].				"Other even numbers are not prime."
+
+	"Now check divisibility by all odd dividers upto the square root of self."
+	maxDivider := self sqrt asInteger + 1.
```
blairmcg: One other thought: `#sqrt` is inexact for integers that are too large to be represented accurately as a 64-bit Float, i.e. any integer > `16r20000000000000`. I'm not sure exactly what effect that might have here without further thought/testing.

Core/Object Arts/Dolphin/Base/IntegerTest.cls
```diff
@@ -68,6 +68,14 @@ testIsPowerOfTwo
 	powersOfTwo do: [:each | self assert: each isPowerOfTwo].
 	nonPowersOfTwo do: [:each | self deny: each isPowerOfTwo]!
 
+testIsPrime
+	| isPrimeResults |
+
+	isPrimeResults := #( 1 2 3 4 5 6 7 8 9 10 11 12 13 1000 7919 ) collect: [ :num | num isPrime ].
```
blairmcg: -2, -1, and 0 as well. Some large integer cases would be nice (although these will be increasingly slow)

FunctionPoint:
Hi Blair, I've implemented all your suggestions,
including using an algorithm that calculates the sqrt using only integers (Integer>>sqrtFloor).
Two questions still:
- In what way should I raise an error when trying to sqrt a negative integer?
- (On Git, am newbe): How should I resubmit the updated code? ("squash changes"?)

blairmcg:
@FunctionPoint, excellent, for the negative case, since we don't have a representation for complex numbers, I would just use something like:
```
self < 0 ifTrue: [^self error: 'Unable to represent square root of negative integer'].
```
To add changes to your PR, just commit the new changes, and push them again. GitHub will automatically add them to a new iteration of the PR. I can squash them together later when merging to master if that is appropriate.

blairmcg:
Closing this as it has been open since July. Please feel free to resubmit.

PR #421, opened 2017-07-19
Avoid hang in Windows 7

jgfoster:
Change effective value of Win32Constants QS_ALLINPUT (see https://github.com/dolphinsmalltalk/DolphinVM/issues/135). This solved my immediate problem, though I welcome input on the approach.


blairmcg:
I think the approach is sound. There are a few details I'd like to change, however. Mostly minor cosmetics, but we should use the correct constant for Windows 7 so that any similar change in the VM matches. I'll push some changes to the branch.

Issue #417, opened 2017-07-17
New image command line option (-i) saves with pending UI events that error later


jgfoster:
I have a script that generate a DPRO image with various extra windows open (including an IdeaSpaceShell with a SystemBrowser). That script works fine and opening the saved image works fine. I have a second script that opens the DPRO image, saves it with a new name (-i JadeDev), and then exits (-x). That script finishes without error but subsequent attempts to open the second image almost always results in an error (that can be dismissed without further harm). 

The error is the same each time a particular image is reopened, but will often change when the second script is rerun. The stack shows a UI event sent to an invalid widget (e.g., a null or 16r0 handle) or an attempt to call a function in a library (DLL) that doesn't exist (see below for a Canvas with a null handle). Since the error is the same each time the same image is reopened, I suspect that the pending event is in the queue at the time of the image save. 

It seems that we should wait till the event queue is empty and the system is quiet before the save. We could also view this as a startup problem in which saved events should be treated with some skepticism. 

2:11:05 PM, Monday, July 17, 2017: 'The specified procedure could not be found. (16r7F: The specified procedure could not be found.)'
UserLibrary(ExternalLibrary)>>invalidCall
UserLibrary>>endPaint:lpPaint:
[] in TabViewXP(View)>>basicPaint:wParam:lParam:
[] in Win32Error(Exception)>>_evaluateHandler:in:
BlockClosure>>ifCurtailed:
BlockClosure>>ensure:
Win32Error(Exception)>>_evaluateHandler:in:
Win32Error(Exception)>>_propagateFrom:
Win32Error(Exception)>>_propagate
Win32Error(Exception)>>signal
Win32Error class(Exception class)>>signal
GDILibrary(ExternalLibrary)>>systemError
Canvas {16r0} >>textExtent:
TabViewXP>>paintVerticalLabel:on:in:
TabViewXP>>paintVerticalTabLabelsOn:
TabViewXP>>onPaintRequired:
[] in TabViewXP(View)>>basicPaint:wParam:lParam:
ExceptionHandler(ExceptionHandlerAbstract)>>markAndTry
[] in ExceptionHandler(ExceptionHandlerAbstract)>>try:
BlockClosure>>ifCurtailed:
BlockClosure>>ensure:
ExceptionHandler(ExceptionHandlerAbstract)>>try:
BlockClosure>>on:do:
TabViewXP(View)>>basicPaint:wParam:lParam:
[...]

PR #423, opened 2017-07-20
Avoid running certain pending actions on image startup

jgfoster:
This addresses #417 and uses image start time (instead of a GUID) to identify command line tasks that should not carry across to a new session. I'm happy to have a different approach; this is along the lines (though not exact) with the earlier discussion.


jgfoster:
While I didn't have a strong objection to the `GUID new` idea, it seemed a bit more complicated to add an instance variable and make sure it was set properly and startTime is managed by the OS and provide a greater utility (not just a random string, but otherwise useful information that an application might desire). But I agree that it added several methods to several classes, so was not an isolated solution.

I have a new commit that uses `imageVersionMinor`, but the final version is a bit more complex than I had hoped. Since the counter gets incremented on image save rather than startup, any action that might cause a save can prevent subsequent actions unless we update the cached value. Let me know your thoughts!

[For what it's worth, I've refactored my application build process to avoid the exit step, so while I enjoy working on this, I'm not waiting for a solution!]

blairmcg:
Yes, I would agree that the information provided by the GetProcessTimes could indeed be useful. It would be useful to have, but perhaps not at this low-level in the system since our objective is to try and cut down the content of the Base package over time. While a Guid is just a random identifier, there are some other potential uses of a statistically (globally) unique session identifier, for example it could be used to correlate telemetry or error reports without needing PII.

Are you happy with the solution you have using the `imageVersionMinor`? Your comment as to the interference of one action with the next concerns me a bit. I'd not thought particularly deeply about the requirement to have an id that didn't change in the session. If you are happy with this solution we can go with it, but it really wouldn't be difficult to add an inst var to SessionManager and initialize that with a Guid at an appropriate early stage of startup. Or alternatively we could store the image sequence number in an inst var in the same way, and then compare the value captured at the time the deferred action blocks are created against the current value of the SessionManager inst var (rather than the current value from the image header, which might have been updated by a subsequent save within the same session).

jgfoster:
I decided to try the GUID approach (before criticizing it further!) and it turns out to be rather simple. I'm quite happy with it. Let me know if you think it needs anything further (e.g., I didn't add an accessor for the new ivar).

blairmcg:
That seems good. I think we should call it `sessionId` though, and an accessor would be nice (though not needed at this point).

PR #429, opened 2017-07-27
Virtual Categories for Loose Methods

rko281:
When working with extended classes (classes with loose methods) it can be difficult to identify the methods defined by a particular class. This change adds a virtual method category for each package in the image, which group together loose methods defined by individual packages. This allows the methods defined by a particular package to be quickly identified in the browsers.

### DIFF
--- a/Core/Object Arts/Dolphin/IDE/Base/ClassBrowserAbstract.cls
+++ b/Core/Object Arts/Dolphin/IDE/Base/ClassBrowserAbstract.cls
@@ -114,7 +114,7 @@ actualClassChainMethods
 	^dict values!
 
 addCategoriesOfMethod: aCompiledMethod
-	(aCompiledMethod realCategories difference: categoriesPresenter model asSet) 
+	(aCompiledMethod categories difference: categoriesPresenter model asSet) 
 		do: [:each | categoriesPresenter model addCategory: each]!
 
 addMethod: method toCategory: target 
@@ -1577,7 +1577,7 @@ onTipTextRequired: tool
 		ifTrue: 
 			[| cat |
 			cat := self category.
-			^cat == self allCategory 
+			^(cat == self allCategory or: [cat acceptsAdditions not])
 				ifTrue: ['New unclassified method']
 				ifFalse: ['New method in <1p>' expandMacrosWith: cat name]].
 	(cmd == #historyBack and: [history hasPast]) 
--- a/Core/Object Arts/Dolphin/IDE/Professional/SystemBrowserShell.cls
+++ b/Core/Object Arts/Dolphin/IDE/Professional/SystemBrowserShell.cls
@@ -135,7 +135,22 @@ selectionEnvironment
 		ifFalse: [classEnvironment]!
 
 slideyPinNames
-	^super slideyPinNames , #('packagesSlidey')! !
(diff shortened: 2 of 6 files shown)


Core/Object Arts/Dolphin/IDE/Base/ClassBrowserAbstract.cls
```diff
@@ -114,7 +114,7 @@ actualClassChainMethods
 	^dict values!
 
 addCategoriesOfMethod: aCompiledMethod
-	(aCompiledMethod realCategories difference: categoriesPresenter model asSet) 
+	(aCompiledMethod categories difference: categoriesPresenter model asSet)
```
blairmcg: CompiledMethod>>categories is slow because it evaluates all the virtual categories. This method is evaluated in response to new methods being compiled into the system (or removed or categorised), and so slowing it down is going to have a big impact on operations such as performing file ins with browsers open. This is why the browsers haven't dynamically changed the set of virtual categories shown in the past. This needs to somehow skip categories which return true to `#isSlowToEnumerate`.
Either that or don't worry about the fact that the virtual categories may get stale.

Core/Object Arts/Dolphin/IDE/Professional/SystemBrowserShell.cls
```diff
@@ -135,7 +135,22 @@ selectionEnvironment
...
+
+	| class package |
+
+	class := self actualClass.
+	self hasPackageSelected ifTrue: [package := self packages anyOne].
+	(class notNil and: [package notNil and: [package includesLooseMethodsFor: class]]) ifFalse: [^super updateCategories].
+
+	categoriesPresenter model list: self actualClassChainCategories asOrderedCollection.
+	categoriesPresenter expand: self allCategory.
+	(LooseMethodCategory forPackage: package)
```
blairmcg: It seems undesirable architecturally to have a dependency on a particular type of virtual category in the browser. Although I think the outcome of displaying just the loose methods when a class with loose methods in the package is selected can be useful, I'm not convinced it should be the default, and I don't like the fact that changing class selection automatically makes a selection in the category tree.  I think it might be better to achieve similar functionality by just adding an option to show only owned methods of classes (which could be toggled on and off like the "inherited methods" filter), especially as this could also work for multiple package selections.

rko281:
Thanks for the feedback and analysis. 

Optimising as much as is readily possible (without making wider changes to the system) I've reduced the relative slowdown between realCategories and categories from ~25x to ~8x. As such I think performance is still going to be unreasonable when filing in etc. so I'll need to take a different approach.

One option I'll investigate is to only do a full category refresh (i.e. using categories) when selecting a class (and possibly when a method is repackaged), using the simpler realCategories refresh when methods are added/removed/categorized. Another option is to drop the categories idea and instead add an additional panel to the Categories/Protocols/Variables section to filter methods by Package.

> I think it might be better to achieve similar functionality by just adding an option to show only owned methods of classes (which could be toggled on and off like the "inherited methods" filter), especially as this could also work for multiple package selections.

Ironically I had such a feature in my old Tools+ package for D5; the loose method categories was an attempt to give similar functionality using the standard Dolphin mechanisms. So I may possibly go back to just a simple filter...

PR #451, opened 2017-08-11
Inno Setup Script refactor

rejurime:
The way that was use to set the setup properties was change using defines without changing the file names, the paths and the version.
Add License and icons in the wizard.


blairmcg:
I don't have any particularly strong opinions about this, but tried a test build of the DolphinRelease definition, and it ran successfully and produced a setup.exe that installed successfully. So in it goes...

PR #478, opened 2017-11-06
Fix Class Rename Refactor with Resources

cDemersMitSci:
Fix for class rename refactor not changing class references in view resource code.  I added a replace for 'Smalltalk.' , className . This now changes class references in view resources.  

It seems impure to hard code the 'Smalltalk.' prefix.  I think this notation was a starting point for potentially supporting multiple namespaces but I could not find an easy way to get a classes namespace.  Someone with a broader understanding may be able to improve the elegance of this fix.

### DIFF
--- a/Core/Contributions/Refactory/Refactoring Browser/Refactorings/RenameClassRefactoring.cls
+++ b/Core/Contributions/Refactory/Refactoring Browser/Refactorings/RenameClassRefactoring.cls
@@ -26,6 +26,7 @@ renameReferences
 	| replacer |
 	replacer := (ParseTreeRewriter replaceLiteral: className with: newName)
 				replace: className with: newName;
+				replace: 'Smalltalk.' , className with: 'Smalltalk.' , newName;
 				replaceArgument: newName
 					withValueFrom: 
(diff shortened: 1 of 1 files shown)


blairmcg:
Thanks for the fix. You could send the environment message to the class to get the containing name scope, but this is also hard coded in the literal filer anyway.

blairmcg:
Oh, and regarding line endings, this can be tricky to get right in git because it runs across platforms and there is no ideal solution for line endings that suits all situations. By default Git stores LFs (IIRC) in the repo, and the translates text files to/from the standard line ending convention of the host OS when checking out. Dolphin only runs on Windows, however, and doesn’t handle other conventions entirely transparently, so we have configured the repo to always use crlf for Smalltalk source files. Usually this means you won’t encounter the problems you saw. However some files in the repo still have the wrong line endings from changes made before the repo was so configured. Not sure if that applies in this case. The only other reason I can think of would be because you somehow have local configuration that is overriding the setting in the root .gitattributes file, or because you are using a tool that is doing its own thing.

Regards

Blair

Issue #499, opened 2018-02-04
Support UTF8 encoded international strings in code editing tools


blairmcg:
Scintilla can display arbitrary Unicode code points from UTF8 encoded content. Although there are many issues with using UTF8 encodings in a system originally designed for 8bit fixed width characters, enabling the display of UTF8 encoded content will at the very least make it easier to work on support for wide characters.

blairmcg:
Addressed in 7.1, which provides full support for UTF encoded strings. Although this obviously doesn't address all internationalization issues, it is an important building block.

PR #500, opened 2018-02-05
Support UTF8 encoded international strings in code editing tools #499 

blairmcg:
* Configure all ScintillaView resources to utf-8 code page.
* Perform wide<->multibyte conversions to/from UTF-8
* Add simple unit test for conversions

Note there is a small risk of application impact from this change since the assumed code page of all byte string<->wide string conversions in the image is now CP_UTF8, i.e. that the narrow strings are UTF8 encoded. Previously they were treated as being encoded per the so-called ANSI code page, CP_ACP. This code page has some encodings (notably for the copyright character) that are not valid in UTF-8.
Given that the offending copyright character was used in the version resources of the Dolphin stubs, there may be issues with application deployment, or other cases where strings containing invalid encodings are converted, especially if round-tripped, since the encoding will be lost. At present the conversion will not fail, but will replace the invalid characters with a special marker character, which might then cause visual impact in the application. An example can be seen on the Dolphin splash screen:

image

The fix for this will be made in a new VM version, but there may be other cases so its worth keeping an eye out, particularly in deployed applications. 

I think the change is of sufficient important to warrant overriding the compatibility concerns, especially as these should be easily fixable by adjusting the offending text/strings. It enables the use of international characters in workspaces. For example:

image

Most controls, however, will still not display the strings correctly. This is because the ANSI apis and control classes are used, and unfortunately Microsoft chose not to support UTF8 through the ANSI apis/controls when they introduced the CP_UTF8 code page in Windows 7 (perhaps rightly, as this would have broken backwards compat). This can only really be fixed by switching to the wide character (W-suffix) APIs and control classes as these accept UTF-16 encoded strings, which is considered the "native" format on Windows. Here is an example with GDI+, which only has W versions and so can already be made to work already:

image


blairmcg:
One further note of caution: There are many other open questions raised by this change, mainly arising from adopting a variable length encoding in a class hierarchy designed way-back in the early days of Smalltalk with a fixed encoding in mind. For example the `#size` reported for a UTF8 string will be the number of bytes, not the number of characters. Similarly when a string is enumerated, the individual bytes will be enumerated, some of which will not be valid characters but rather lead bytes of multi-byte character encodings.  I have opinions on some of these issues as to the right solutions (and the [UTF8 everywhere manifesto ](http://utf8everywhere.org/) has had some input into this opinion forming), but they are just opinions. For example I think it may be reasonable that `#size` reports the number of "code units" (in this case, bytes). On the other hand it seems to me that the enumerators such as `#do:` should enumerate the actual characters after decoding into Unicode code points. I'm aware that there is some inconsistency here, but as the manifesto says "a string length function should not necessarily return one for the string ‘🐨’ to be considered Unicode-compliant." (and yes, you can paste that Koala character into a Dolphin workspace after this change :-)).
So anyway, this change leaves many loose ends. I apologise if you run into those, but if you do please report them. If it turns out to be too disruptive, I will back it out and re-think.

blairmcg:
I've come to the conclusion that it is too impactful to just make the assumption that all Strings are UTF8 when the previous assumption was that ANSI. Although this certainly doesn't make handling of characters that require multibyte encodings any worse (they didn't work, after all), it does mean that a lot of pre-existing text will now be interpreted differently (basically anything with byte characters outside the ASCII range, e.g. être. There is less impact than I might have expected from this from tools perspective, in that the streaming in and out of code represented as UTF8 encoded strings to files, etc, works fine. The problem is that any pre-existing strings containing non-ASCII characters might be wrongly interpreted. Worse than that, the same would apply to data files.
The issue is really with changing the encoding for all Strings to be UTF8. This breaks backwards compatibility too much (as well as a load of String manipulation methods, e.g. `asUppercase`, `reverse`, although those could be fixed). I think we need to make String itself abstract, and then introduce Ansi and UTF8 subclasses. For new stuff UTF8 encoding would be used with the appropriate BOM being emitted. Stream classes would need to be enhanced to support different text encodings for input and output in some way.
This was an interesting experiment, but I'm going to back it out.

PR #507, opened 2018-02-20
Always pass a UTF16 string to SetWindowText

blairmcg:
Windows will translate automatically for non-Unicode windows. This will
therefore work for Unicode Windows, and with any of Ansi, UTF8 or UTF16
encoded string values.

@jgfoster, this should fix your issue with populating text controls from a IXMLDomNode without needing explicit `#asAnsiString` conversions.


jgfoster:
Yes, that fixed it. Thanks very much.

Issue #514, opened 2018-03-07
Process does not understand #webHttpClientSocketToClose


Debiller777:
So HttpClient class has class method "getUrl", however when I call it I get error that Process doesnt understand method "webHttpClientSocketToClose". So I decided to browse it. My try was unsuccessful. Then I decided to browse all references to this method. It turned out that Process>>itcSafeTerminate used 
this method and sent it so self, though it wasnt implemented. Also it it used another undefined method "asWebString". So, what should I do? Where can I find sources of these methods.

blairmcg:
The package is a third party contribution that hasn't been imported completely, or is missing dependencies. Given the absence of tests, there is little guarantee it will work even with the missing pieces. Someone will need to volunteer to fix it, or I would be happy to remove it.

I'm not sure how practical it is, or how much use it has received, but there is a wrapper around the Microsoft HTTP client through IWinHttpRequest, e.g

```
http := IWinHttpRequest new.
http open: 'get' url: 'http://www.google.com'.
http send.
http responseText.
```

It needs some work on convenience methods to make it more usable. Others may have alternative suggestions.

Debiller777:
Woah thank you. That was very helpful. Where can I find this Http client?

ср, 7 марта 2018 г., 20:33 Blair McGlashan <notifications@github.com>:

fxgallego:
@Debiller777  ``IWinHttpRequest`` is bundled with Dolphin. Use the find class tool.

Debiller777:
@fxgallego Thanks you!

blairmcg:
The HTTPClient has been removed from the standard build

Issue #496, opened 2018-01-28
HttpClient doesn't works


fxgallego:
HttpClient getUrl: 'http://www.google.com/' gives me a message not understood because it expects #webHttpClientSocketToClose: to be defined in Process. It appears that some Process extensions are lost.

This applies to EmailSMTPConnector as well.

BTW: I think that ITC Gorisek packages define too much extensions and not always is clear why we need them, for example Timestamp. IMO STS shouldn't be loaded by default this days.

blairmcg:
The HTTPClient has been removed from the standard build

PR #517, opened 2018-05-09
Add issue #514 by removing the broken HTTP Client and dependent packages

blairmcg:
If someone wants to fix this, feel free to add it back to the product
contents. Otherwise I suggest enriching the Win HTTP Client wrapper.


blairmcg:
Also addresses #496

PR #526, opened 2018-05-17
SHA1 BCrypt update

FunctionPoint:
Small update to process changed (now standard) hash byte order output from DolphinSureCryptoLib in the VM solution. See commit there.

### DIFF
--- a/Core/Object Arts/Dolphin/DolphinSure/SecureHashAlgorithm.cls
+++ b/Core/Object Arts/Dolphin/DolphinSure/SecureHashAlgorithm.cls
@@ -21,11 +21,12 @@ finalHash
 	"Private - Answers the final hash from the context and closes it"
 
 	| totals hash |
-	totals := DWORDArray new: 5.
+	totals := ByteArray new: 20.
 	self shaLibrary shaGetHash: self context lpHashBytes: totals.
 	self resetContext.
+
 	hash := 0.
-	1 to: 5 do: [:i | hash := hash bitOr: ((totals at: i) bitShift: 32 * (5 - i))].
+	1 to: totals size do: [ :i | hash := ( hash bitShift: 8 ) bitOr: ( totals at: i ) ].
 	^hash!
 
 hashIn: aByteArray 


blairmcg:
In order for this to pass validation you will need to make a commit to update the VM submodule. Git should show a status change if you have a correctly initialised repo (`git submodule update --init` has been run) and you have made a commit in the submodule. You then need to add Core\DolphinVM and if you have pushed the VM repo with the corresponding change, the CI build will be able to clone the change in the submodule and the build should pass. This needs to be done before it can be merged or the build will continue to fail.

FunctionPoint:
Ah, thanks for the input. I'm on a weekend away now, it might be Monday or Tuesday to do this. Hopefully this doesn't interfere with other developments..

FunctionPoint:
Okay, I may need a litte jump-start with this Git stuff (my excuse is limited time :)  
.
I went to my Dolphin (not DolphinVM) development folder and did: 
    git submodule status
Indeed, it shows Core/DolphinVM as an existing submodule.
Then form that location I did: 
     git submodule update --init
Some disk activity, no message, and a Core/DolphinVM folder is there with a clone of the VM.
.
How to proceed now:
"You then need to add Core\DolphinVM" to what, how?
"and if you have pushed the VM repo with the corresponding change" make the same change in the submodule? Seems a re-use of the change in ../DolphinVM should be possible somehow.
.
Otherwise:
I did simple test file change (added a comment), but that shows up in GitHub desktop as a "dirty" subproject commit. Not as the usual diff.
For the DolphinVM folder, I also did a PR. Was that okay?
.
Some tips would be greatly appreciated.

blairmcg:
Well in theory the process when working on changes that require changes in both the image and VM is as follows:

1. You makes your changes in both repos in your clones where the VM is a submodule of the main Dolphin repo,until you are happy (e.g. local boot and test run work).

1. You commit your VM change.
1. In the main repo you will note that `git status` reports there are changes in the submodule; it will say something like `modified:   Core/DolphinVM (new commits)`. You now need to make a commit in the Dolphin repo that will update the pointer in the main repo to reference the new VM commits. To do this just `git add Core/dolphinVM` in the Dolphin repo and commit.
1. If you've not done so already, you push your VM repo. Ignore the DolphinVM CI build failure - it won't have the corresponding images changes.
1. Now you can push your Dolphin repo. 
1. When the CI runs, despite that the VM commit is not in master, it gets picked up because of the submodule reference to the correct SHA.

That's the theory, but it strikes me there may be a snag. This does work if one can push to the main repo, but likely not in a fork, unless github have got something clever in place, we'll see. It looks like you also need to merge upstream changes (preferably by rebasing your branch onto the current master, and fixing any merge conflicts, since this keeps a single line history which is easier to work with later). If you have problems with this, let me know and I can probably patch it up for you.

In hindsight, the original decision to split the repos probably wasn't the right one. Converting to a submodule was easy because of the way it was originally set up to be nested, but although it is now possible to make synchronised changes, it is still rather awkward. I think it will be necessary to merge down to one repo to allow for a sensible development process that is not too onerous.

FunctionPoint:
Thanks forthe guide, Below is how far I came, annotated with "> ".
PS The starting situation is that I've also committed the VMchange in the "dolphinsmalltalk/DolphinVM" repo, so not in the submodule.

Well in theory the process when working on changes that require changes in both the image and VM is as follows:

1.      You makes your changes in both repos in your clones where the VM is a submodule of the main Dolphin repo,until you are happy (e.g. local boot and test run work).
> Check. Boot and regressions tests run, including SecureHashAlgorithmTest, that hits the modified code.

2.      You commit your VM change.
> Check. Reports changed files correctly.   (But had to use command-line git iso GitHub desktop ... )

3.      In the main repo you will note that git status reports there are changes in the submodule; it will say something like modified: Core/DolphinVM (new commits).
You now need to make a commit in the Dolphin repo that will update the pointer in the main repo to reference the new VM commits.
To do this just git add Core/dolphinVM in the Dolphin repo and commit.
> Check. Reports commit in subproject.

4.      If you've not done so already, you push your VM repo.
Ignore the DolphinVM CI build failure - it won't have the corresponding images changes.
> So this is the non-submodule repo? Did this last week with the same changes. (see 1st line)

5.      Now you can push your Dolphin repo.
> Do you mean, create a new Pull Request?
> When do this form GitHub in the browser, is just shows my original pull request last week with the changed method "SecureHashAlgorithm>>finalHash"
> No mention of changes in submodules.
>  Not sure how to proceed here...

6.      When the CI runs, despite that the VM commit is not in master, it gets picked up because of the submodule reference to the correct SHA.

That's the theory, but it strikes me there may be a snag. This does work if one can push to the main repo, but likely not in a fork, unless github have got something clever in place, we'll see. It looks like you also need to merge upstream changes (preferably by rebasing your branch onto the current master, and fixing any merge conflicts, since this keeps a single line history which is easier to work with later). If you have problems with this, let me know and I can probably patch it up for you.

> Yesss, if you could please merge my changes in your own release process, I would be very grateful.
> I'm pretty confident the changes are okay. They also test well.

In hindsight, the original decision to split the repos probably wasn't the right one. Converting to a submodule was easy because of the way it was originally set up to be nested, but although it is now possible to make synchronised changes, it is still rather awkward. I think it will be necessary to merge down to one repo to allow for a sensible development process that is not too onerous.
> I totally agree.
> Probably, if I study Git for a day or two, I'll get the submodule process under control.
> But I'd rather code than wrestle with Git. (I liked SourceSafe :-) And I have plans for more VM / image changes.
> Also other newbies wil have the same learning curve.
>
> So..., why don't we just move the VM to a subfolder of the Dolphin project in Git, and completely *remove* the separate DolphinVM Git project.
> Commits that require parallel changes in the VM and the image then become easy again.
> When newcomers don't want to build the VM with Visual Studio, that's fine, they can still use the FetchVM script.
> You will always get the VM source code when you do a clone, but diskspace isn't a problem these days. It will just be unused for some.
>
> Is that an option for you?

FunctionPoint:
PS git status reports "modified:   Core/DolphinVM (untracked content)".
So locally DolphinVM wasn't added properly added as a submodule.
That's probably why I don't see the VM change also in the PR for Dolphin.
There seems to be a procedure to correct this, but rather start fresh and have this change pusher through somehow...

blairmcg:
Making the VM submodule into a subfolder was what I meant when I said "merge down to one repo", just using the word merge to describe the process of actually merging it in to preserve the history, which I will do when I have time to look into how to achieve that.

I don't remember it being at all difficult to fix up the nested repo into being recognised as a submodule. If the git submodule update --init didn't just work, I would have just googled it and (doubtless) found some instruction on stackoverflow. However it would also be easy to just create a fresh clone of your fork, do the git submodule init, and then checking out the VM submodule to the desired commit, which will put you at the right point.

I can't fix this PR, because I can't push to the repo/branch you PR'd from.  However I pushed a fixed branch to the main Dolphin repo, so if you just fetch from upstream into your fork and then reset your master to functionpoint/sha1bcrypt (95eb7128) and re-push back to your fork, this PR should get fixed up. The build ran or the branch ran green anyway, and is picking up your change to the DolphinSure dll: https://ci.appveyor.com/project/dolphinsmalltalk/dolphin-db22v/build/7.1.0.82

FunctionPoint:
After looking at this for a  while again: (okay, admittedly only for an hour or 2)
This has become too messy with git submodules on my end.
I'm closing this pull request, also since it was only the suboptimal solution.
I'll wil start fresh with a new fork and focus on the Dolphin solution that does not use DolphinSureCrypto.dll at all anymore. 
This solution additionally has the advantage that it can be comitted seperately to Dolphin and the VM, 
without breaking the CI.
.
(I hope you find the time soon to add the VM as a sub*folder* of the Dolphin repo, 
so submodules are no longer needed. Is preserving the history really important?
We could just ceep the current separate VM repo around for a while "read-only".)

FunctionPoint:
PS I tried a fresh clone just now with GitHub Desktop.
Near the and of the clone, when it's busy with the submodule: Core/DolphinVM,
I get the following error: "A submodule points to a commit which does not exist."
.
Something on my side, or not...?

FunctionPoint:
Hmmm, this problem persists.
With a new computer, I have:
- On Github deleted the forked repositiory: FunctionPoint/Dolphin.
- On Github created a new fork FunctionPoint/Dolphin from DolphinSmalltalk/Dolphin.
- On local machine deleted all repositories from GitHub Desktop and from disk.
- Created a new clone from repo: FunctionPoint/Dolphin.
Then I get the following error: "A submodule points to a commit which does not exist."
.
Cloning from DolphinSmalltalk/Dolphin directly works OK.
.
Any suggestions?

blairmcg:
No, but you'll probably get better diagnostics from the git command line.
I'll look at merging in the submodule now that there is nothing outstanding using that structure.

FunctionPoint:
Yep, I'd though I'd try that. 
The command line git does not give any errors, see below.
Well okay, I could go and use that for a while then... 
--
git clone git://github.com/functionpoint/dolphin
Cloning into 'dolphin'...
remote: Counting objects: 23888, done.
remote: Compressing objects: 100% (107/107), done.
remote: Total 23888 (delta 45), reused 97 (delta 39), pack-reused 23742
Receiving objects: 100% (23888/23888), 75.78 MiB | 621.00 KiB/s, done.
Resolving deltas: 100% (16067/16067), done.
Downloading Core/Object Arts/Dolphin/Base/Boot/DBOOT.img7 (1.30 MB)
Downloading Core/Object Arts/Dolphin/Base/Boot/DBOOT.sml (88 B)
Downloading DBOOT.img7 (1.37 MB)190)
Downloading DBOOT.sml (2.31 MB)4190)
Checking out files: 100% (4190/4190), done.

FunctionPoint:
... but the clone above does not get the summodule.
The command line to fetch the submodule gives the error of the offending object:
----
git submodule update
Cloning into 'D:/Richard/Development/GitHub/Dolphin/Core/DolphinVM'...
error: Server does not allow request for unadvertised object c4095e0118ae77e47ddf0b2f2b09846dd684f5d8
Fetched in submodule path 'Core/DolphinVM', but it did not contain c4095e0118ae77e47ddf0b2f2b09846dd684f5d8. Direct fetching of that commit failed.

PR #530, opened 2018-06-14
Merge the VM repo into the main repo

blairmcg:
The proposal here is to merge the VM repo into the main Dolphin repo. Why?

-  Separate repo's make life very difficult when making changes to the VM that require changes to the image, and vice versa. One has to jump through hoops in order to keep the CIs running, including on occasion having to write temporary compatibility code in the VM and certainly in the tests. Frankly it would have been impossible to make a change as significant as the recent Unicode enablement with this arrangement.
- The interim approach of using a submodule for the VM does make it possible to make synchronised changes (and enabled the Unicode changes in 7.1 to be merged relatively easily), but involves more ceremony than one might like, and is tricky to handle from a fork.
- Releases of Dolphin itself have been sporadic - much less frequent than VM releases - as it involves an extra step that I keep forgetting to do.

The best solution at this point would seem to be to merge down to one repo. This need not necessarily mean that those who don't want to build the VM need do so. We can keep publishing the VM builds.

Once this goes in the existing DolphinVM repo will be dead and any forks similarly. If you have pending VM work, you'll need to merge it over, although this is possible in automated fashion since this PR merges the commits together in a way that allows git to track changes between the two. All of the original commits are preserved with their original SHAs and git blame and other history commands will follow through normally without needing to do anything special.

I will in time take down the VM builds and update the VM repo readme to indicate its state. The Dolphin release build will be modified to publish the VM binaries and the fetchvm script updated accordingly.

I'll leave this for a day or two to allow for complaints or better suggestions. In the meantime it will be possible to base off the blairmcg/mergevm branch to test it out or do new work.


blairmcg:
In the absence of comments to the contrary, it's going in...

FunctionPoint:
Ah didn't see this sooner to comment: Go for it 👍

Issue #568, opened 2018-06-30
A Scintilla upgrade is now due


blairmcg:
We are currently on Scintilla 3.7.5, which is from May 2017. The current version is 4.1. Although it made sense to remain on 3.7 while the new major release settled out, now that 4.1 is out it is time to catch up and add support for any new features we want to expose in the class wrapper.

PR #572, opened 2018-07-02
A Scintilla upgrade is now due #568

blairmcg:


Core/Object Arts/Dolphin/IDE/Base/SmalltalkWorkspace.cls
```diff
@@ -66,7 +66,6 @@ applyOptions
 	view autoCompletionSeparator: $\x7F.
 	self areVariableTipsEnabled ifTrue: [view isBackgroundDwellEnabled: true].
 	"#2137"
-	view sciSetCodePage: KernelLibrary default getACP.
```
blairmcg: Not part of the upgrade. This should have been removed with the switch to UTF-8, and may have been causing problems for anyone using a code page that Scintilla recognised.

PR #594, opened 2018-07-20
Fix #593 Some HTTP_PROPERTY_FLAGS subclasses doesn't implement #propertyId

### LINKED ISSUE #589: Consider adding SequenceableCollection#indexOf:startingAt:
This method was previously defined in the "Dialect Abstraction Layer" package, now removed, and looking at other methods similar to this seems that `#indexOf:startingAt:` is missing. (See indexOfAnyOf:, indexOfSubCollection: families)

```
indexOf: target startingAt: start
	"Answer the <integer> index of the first element of the receiver which is 
	equal to the <Object> argument, target, within the receiver. If the receiver 
	does not contain any elements equal to target, answer 0."

	^self nextIndexOf: target from: start to: self size
```

### LINKED ISSUE #593: Some HTTP_PROPERTY_FLAGS subclasses doesn't implement #propertyId
`#propertyId` is only defined for a couple of classes, probably those that are used in examples.
In addition to this HTTP_LOGGING_INFO needs instance variables to hold DirectoryName and SoftwareName strings. The length passed as parameter for both fields should be the size in bytes of the string without the null terminator.
I'll provide a PR to address this.

fxgallego:
Fix #593
Adds several #propertyId definitions
Adds instance variables for HTTP_LOGGING_INFO
Modifies DirectoryName and SoftwareName setters to ensure UTF16 strings are passed and its lengths are correctly set.
Adds a new example, `HttpApiLibrary>>example8` that currently doesn't works.

Fix #589

### DIFF
--- a/Core/Object Arts/Dolphin/Base/SequenceableCollection.cls
+++ b/Core/Object Arts/Dolphin/Base/SequenceableCollection.cls
@@ -724,6 +724,13 @@ indexOf: target ifAbsent: exceptionHandler
 		ifTrue: [exceptionHandler value]
 		ifFalse: [index]!
 
+indexOf: target startingAt: start
+	"Answer the <integer> index of the first element of the receiver which is 
+	equal to the <Object> argument, target, within the receiver. If the receiver 
+	does not contain any elements equal to target, answer 0."
+
+	^self nextIndexOf: target from: start to: self size!
+
 indexOfAnyOf: aCollection startingAt: anInteger 
 	"Answer the one-based integer index of the first encountered element of the receiver which 
 	is equal to one of the elements of the <Collection> argument, starting from the specified
@@ -1265,6 +1272,7 @@ writeStream
 !SequenceableCollection categoriesFor: #includesKey:!public!testing! !
 !SequenceableCollection categoriesFor: #indexOf:!public!searching! !
 !SequenceableCollection categoriesFor: #indexOf:ifAbsent:!public!searching! !
+!SequenceableCollection categoriesFor: #indexOf:startingAt:!public!searching! !
 !SequenceableCollection categoriesFor: #indexOfAnyOf:startingAt:!public!searching! !
 !SequenceableCollection categoriesFor: #indexOfSubCollection:!public!searching! !
 !SequenceableCollection categoriesFor: #indexOfSubCollection:startingAt:!public!searching! !
--- a/Core/Object Arts/Dolphin/System/Win32/HTTP_LISTEN_ENDPOINT_INFO.cls
+++ b/Core/Object Arts/Dolphin/System/Win32/HTTP_LISTEN_ENDPOINT_INFO.cls
@@ -53,6 +53,10 @@ defineFields
 	super defineFields.
 	self
 		defineField: #EnableSharing type: BOOLEANField new offset: 4.
-	self byteSize: 8! !
+	self byteSize: 8!
+
+propertyId
+	^HttpServerListenEndpointProperty! !
 !HTTP_LISTEN_ENDPOINT_INFO class categoriesFor: #defineFields!**auto generated**!initializing!public! !
+!HTTP_LISTEN_ENDPOINT_INFO class categoriesFor: #propertyId!public! !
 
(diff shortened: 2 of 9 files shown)


Core/Object Arts/Dolphin/Base/SequenceableCollection.cls
```diff
@@ -724,6 +724,13 @@ indexOf: target ifAbsent: exceptionHandler
 		ifTrue: [exceptionHandler value]
 		ifFalse: [index]!
 
+indexOf: target startingAt: start
+	"Answer the <integer> index of the first element of the receiver which is 
+	equal to the <Object> argument, target, within the receiver. If the receiver 
+	does not contain any elements equal to target, answer 0."
+
+	^self nextIndexOf: target from: start to: self size!
```
blairmcg: Generally speaking tests should accompany base system additions, but given this is a convenience wrapper on top of `nextIndexOf:from:to:`, we can live without here.

Core/Object Arts/Dolphin/System/Win32/HTTP_LISTEN_ENDPOINT_INFO.cls
```diff
@@ -53,6 +53,10 @@ defineFields
 	super defineFields.
 	self
 		defineField: #EnableSharing type: BOOLEANField new offset: 4.
-	self byteSize: 8! !
+	self byteSize: 8!
+
+propertyId
+	^HttpServerListenEndpointProperty! !
 !HTTP_LISTEN_ENDPOINT_INFO class categoriesFor: #defineFields!**auto generated**!initializing!public! !
+!HTTP_LISTEN_ENDPOINT_INFO class categoriesFor: #propertyId!public! !
```
blairmcg: 'constants' category is missing on all these (and the originals)

Core/Object Arts/Dolphin/System/Win32/HTTP_LOGGING_INFO.cls
```diff
@@ -44,7 +44,9 @@ DirectoryName
 DirectoryName: anUtf16String
 	"Set the receiver's 'DirectoryName' field to the value of the argument, anUtf16String"
 
-	bytes uintPtrAtOffset: ##(self offsetOf: #DirectoryName) put: anUtf16String yourAddress!
+	directoryName := anUtf16String asUtf16String.
```
blairmcg: These are auto-generated methods. They'll get overwritten the next time the structure is compiled out. You could just take them out of the auto-generated category to prevent that, but the normal thing to do would be to add a wrapper method (in this case it would be `directoryName:`) that calls the auto-generated method and adds the extra to perform the conversion and save the string to prevent it being GC'd.
The setting of the length could also be generated automatically if the IDL is better defined. I updated it (see [blairmcg/winhttpservertlb](url) https://github.com/blairmcg/winhttpservertlb) and re-generated this and I got:

```
DirectoryName: anUtf16String
	"Set the receiver's 'DirectoryName' field to the value of the argument, anUtf16String"

	bytes uintPtrAtOffset: ##(self offsetOf: #DirectoryName) put: anUtf16String yourAddress.
	self DirectoryNameLength: anUtf16String size << 1
```
It also regenerates the reader as:
```
DirectoryName
	"Answer the <Utf16String> value of the receiver's 'DirectoryName' field."

	^Utf16String fromAddress: (bytes intPtrAtOffset: ##(self offsetOf: #DirectoryName))
		length: self DirectoryNameLength >> 1
```
The original generated reader is assuming it is a null-terminated string, which is incorrect since from the structure definition it has a counted length and the null terminator cannot be assumed. Very often there will be one due to allocation granularity, but that is just luck really.

If you find other issues with the structure definitions it would be good to try and update the original IDL used to generate them; there is quite a lot of richness that can be expressed using extension attributes including custom ones understood by Dolphin that I added to capture extra metada.

However, since you are the person mainly working with this I'll leave it up to you what you want to do and commit this anyway.

fxgallego: Ok, I see the .tlb is there in appveyor just built. Think I don't need any more.

fxgallego: @blairmcg generating the wrapper using your last tlb shows changes in several files not related to the changes discussed here:

image

Mostly changes boolean parameters for integer and void* by PVOID.

It's ok? Should I commit all this files?

blairmcg: @fxgallego, that seems to be a regression. Let me check out why that is.

Issue #550, opened 2018-06-26
DateAndTime subtraction order is incorrect


rko281:
DateAndTime>>subtractFromDateAndTime: has the order of subtraction the wrong way round giving incorrect results, e.g.

```
| now oneDay |
now := DateAndTime now.
oneDay := Duration days: 1.
(now + oneDay) - now " -1:00:00:00, should be 1:00:00:00 i.e. oneDay"

```
The method has a rather ambiguous comment that may imply the reversed order is expected, however the current implementation gives incorrect results when compared to (e.g.) Pharo. I also have an old D5 version of this class which has the same comment but with the correct ordering.

Reversing the order of subtraction gives the expected behaviour - I'll submit a pull request shortly.

blairmcg:
@rko281, John, are there any other diffs with respect to the D5 version? This makes me wonder about the heritage of the version that is checked in to git.

rko281:
The D5 version I got from the Seaside 3.0 distribution for Dolphin 6 at

http://www.infoil.com.ar/seaside/ 

The package comment in that version states it's from D5 (though the Duration package states it's from D6).

From a quick diff of DateAndTime.cls it looks equivalent to the current D7.1 version (with fix) aside from some minor formatting changes and references to TIMEZONEINFORMATION (which was part of the D5 package) rather than TIME_ZONE_INFORMATION.

blairmcg:
Ok, thanks John.

PR #606, opened 2018-08-29
Repeat fix for #550

rko281:
Issue #550 has regressed, so reapplying the fix from #551.

Whilst looking at this I ported over the DateAndTime unit tests from Pharo but these currently fail due to missing convenience methods (String>>asDateAndTime, Number>>seconds etc.) and differing tolerances for DateAndTime text formats. Porting these tests as-is would probably require significant rework of the DateAndTime package.


blairmcg:
Sorry John, I’m sure that would have been my fault.

PR #610, opened 2018-09-06
Fix profiler division by zero if sample set is empty

blairmcg:
Also replace the Help menu with the standard one (the old one still had
Dolphin Live Update on it). We really need a way to share menu definitions.


blairmcg:
GitHub have kindly disabled LFS on the repo because somehow or other more than 1Gb of LFS data has been downloaded in some unspecified period. Since the usage limit on the LFS is also 1Gb, just one download of that would be enough.
LFS is a good idea, but not like this. Back to binary files in the repo then, sorry.

Issue #612, opened 2018-09-09
Fix RuntimeSessionManager>>deploy:


FunctionPoint:
Dolphin has a nice quick-deployment option of home made apps that does not require image stripping.

E.g.:  NotepadSessionManager deploy: 'MyNotepad.exe' 

But this command now fails just before closing the image because of a missing method "bootInfo".
To fix that, this methods can be added to SessionManager:

bootInfo
	^ nil.

MyNotePad.exe is generated, but when starting it Windows 10 gives the error message:
"This app cannot run on this PC". The usual option of "Run anyway" is not there.
Is some more version info needed in the *.exe? 
Maybe Windows thinks it's a 16-bit app?

blairmcg:
The result is not an executable binary, its an image file, even though it is wrongly saved as a .exe. To quote from the comment: 

```
The image can be run by Dolphin.exe (e.g. by double-clicking the .img file).
Note that because the application is started by Dolphin.exe the first command
line argument will be the image path.
```

Judging from the fact that the examples refer to session manager classes that no longer exist, the image is saved with the wrong file extension, and the bootInfo problem (which can be safely ignored, since this occurs in the continuing session after the image ahs been saved to the file), I'd guess it hasn't been used/tested for a very long time. If one opens the resulting file with Dolphin.exe, it does appear to work though.

FunctionPoint:
Ah yes, that was a bit of sloppy reading by me. :)
Indeed it works if it's saved as an image file.
I think the bootInfo method should still be added to SessionManager prevent the error message. 
I'll create a small PR.

On a side note:
The Notepad example chosen will try to open the first command line argument as a text file.
So it wil open the image name that must be the first argument, 
*if* the image is named anything else than then default "dpro.img7'.

I stumbled on the "deploy:" function while playing with SessionManager classes and an alternative startup method for them that does not require deployment. And indeed got the same behaviour. :)
More later.

FunctionPoint:
That's fine with me too.
As said, I found this bug while experimenting with a *replacement* for this...

blairmcg:
Closing this, since it is now obsolete as a bug. Any replacement can be a separate enhancement issue.

FunctionPoint:
Thanks for the work-around suggestion.
Going to finish something similar first.
(But must admit coding time has been minimal, with long work hours and weekend festivities :)

PR #617, opened 2018-09-13
Fix RuntimeSessionManager-deploy: (issue #612)

FunctionPoint:
Add bootInfo method to SessionMananger that's expected by the deploy: method.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/SessionManager.cls
+++ b/Core/Object Arts/Dolphin/Base/SessionManager.cls
@@ -170,6 +170,10 @@ basicTertiaryStartup
 
 	^self subclassResponsibility!
 
+bootInfo
+	"The default behaviour is no boot info."
+	^ nil.!
+
 clearSessionState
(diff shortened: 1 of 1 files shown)


blairmcg:
I think the right thing to do here is to just remove the deploy: method. It was added originally because it provided a simple way to create a runtime image if one didn't have the pro version with the deployment wizard. It doesn't really offer anything useful now that everyone has access to that, plus it is confusing and has various inconvenient aspects which are described in the comment (like the problem with command line arguments). Image stripping is very fast anyway, particularly if one turns off all the code removal and then you end up with something close to what #deploy: does, except in more usable form.
A high-level method that scripts the image stripper might be a useful replacement, but as it stands I think this method should just be deleted altogether as an alternative to trying to fix something unsatisfactory.

blairmcg:
Based on your comment elsewhere, I'll remove the #deploy: method. The `bootInfo` message shouldn't be sent to or understood by a `RuntimeSessionManager`, so I won't be merging this one.

PR #628, opened 2018-09-20
Fix #627 EditableListView -> #onValueChangedIn:

### LINKED ISSUE #627: EditableListView -> #onValueChangedIn:
We had a problem with the actual implementation of this method because when the value changed we were refreshing the view, so the view was hidden and its activeEditorCoords were set to nil. By the time this method calls for the activeEditorModelIndex again it returns zero and then raised an BoundsError.

We were able to solve it with a workaround in out code, but we thougt it wasn't a bad idea to keep the index in a local variable in case of the editor's view being hidden while the value is being updated.

msobral:
Keep the index in a local variable in case of the editor's view being hidden while the value is being updated.

### DIFF
--- a/Core/Contributions/Solutions Software/EditableListView.cls
+++ b/Core/Contributions/Solutions Software/EditableListView.cls
@@ -678,13 +678,16 @@ onSetFocus
 
 onValueChangedIn: aColumn
 
+	|activeEditorModelIndex|
+	"We keep the index in case of the view being hidden"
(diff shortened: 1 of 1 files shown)


blairmcg:
Seems reasonable to me, but @rko281 as this is your stuff, do you have a view?

rko281:
Looks a sensible change - fine with me.

PR #641, opened 2018-09-30
Fix End of stream error with nextAvailable: on large binary files

### LINKED ISSUE #640: End of stream error with binary FileStream>>nextAvailable:
Reading a binary FileStream with nextAvailable: raises an End of stream error if the file size is greater than the page buffer size:

```
s := FileStream write: 'test' text: false.
s nextPutAll: (ByteArray new: FileStream.PageSize + 1).
s close.

s := FileStream read: 'test' text: false.
s size. 
[s atEnd] whileFalse: [s nextAvailable: 64] "End of stream error"

```

rko281:
Change direct reference to position inst var in `PositionableStream>>basicNextAvailable:` to `position` accessor method so FileStream can adjust based on its paging. 

Added an explicit SUnit for this case rather than extending existing test data since the extra size slows down `testNextAvailableColon` noticeably.

Fixes #640

### DIFF
--- a/Core/Object Arts/Dolphin/Base/FileStreamTest.cls
+++ b/Core/Object Arts/Dolphin/Base/FileStreamTest.cls
@@ -70,6 +70,22 @@ testFileTimes
 	file := filestream file.
 	self assert: file lastWriteTime asInteger >= now asInteger!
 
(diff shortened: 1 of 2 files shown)


blairmcg:
LGTM

Issue #643, opened 2018-10-02
Error deploying image


jgfoster:
{0AAC0744: cf 0AAC0725, sp 0AAC0758, bp 0AAC0740, ip 4, Metaclass(Object)>>doesNotUnderstand:}
	receiver: Win32Error class
	arg[0]: Message selector: selectorsDo: arguments: 1010CEC0

{0AAC0724: cf 0AAC0701, sp 0AAC0738, bp 0AAC071C, ip 26, Win32Error class(ClassDescription)>>resourceSelectorsDo:}
	receiver: Win32Error
	arg[0]: [] @ 9 in ClassDescription>>resourceIdentifiersDo:
	stack temp[0]: 'resource_'

{0AAC0700: cf 0AAC06E1, sp 0AAC0714, bp 0AAC06FC, ip 16, Win32Error class(ClassDescription)>>resourceIdentifiersDo:}
	receiver: Win32Error
	arg[0]: [] @ 61 in ImageStripper>>resourcesForClasses:

{0AAC06E0: cf 0AAC06B1, sp 0AAC06F4, bp 0AAC06D4, ip 79, [] in ImageStripper>>resourcesForClasses:}
	receiver: nil
	arg[0]: Win32Error

PR #644, opened 2018-10-02
Fix for #643.

jgfoster:
the #selectorsDo: method is in an RB package so doesn't exist during the image stripping process. This is a simple fix. I have no strong opinion if it is right, but it works for me!


blairmcg:
I'll merge this, since it fixes the issue, and move the original method later (since the offending change was to speed up the search for resources, and this will slow it down again). 

It occurs to me that we can detect this class of issue on many occasions by using `SmalltalkSystem>>unimplementedSelectorsIn:` (it surfaces up as the _Browse/Unimplemented Selectors_ command in the package browser). Its not 100% foolproof, but it should allow a test to be defined which will catch future mistakes like the one I made which caused this issue. Or at least it will if I also fix #645.

Issue #711, opened 2019-02-25
Issues with resetting of context in ListView column custom draw


rko281:
Resetting of ListView's per-column custom draw introduced in 601b7736508176ef09a0dda9078d418bf7571003 has a couple of issues.

Firstly the reset is dependent on whether the previous column has a customDrawBlock:

`customDrawSubItem: context
...

	"If not the first column, and the previous column has a custom draw block, then we need to reset the custom draw settings as otherwise we'll end up picking up whatever customizations might have been applied to the previous column."
	(idx ~~ 0 and: [(columns at: idx) customDrawBlock notNil]) ifTrue: [context reset].`

Unfortunately this does not take into account the current ordering of columns (as in `ListView>>columnOrder`) and display of subitems appears to be done based on this ordering. As an example, evaluate:

`listView := ((SmalltalkSystem current browseClass: CRTLibrary) view viewNamed: 'methods') presenterConnectionPoint.
listView allColumns first isAutoResize: false; width: 200.
listView addColumn: 
	(ListViewColumn new
		getTextBlock: [ :obj | 'test'];
		isAutoResize: true;
		yourself).`

The strike-through highlighting of deprecated methods bleeds through from the method name to the newly-added column since the method name column is actually the first column in the view's columns collection and so does not precede the new column in the display order.

Secondly, the current implementation of `NMLVCUSTOMDRAW>>reset` only resets the italic and strike-through font emphases. Other emphases (bold, underline etc.) continue to bleed through.  Example:

`((SmalltalkSystem current browseClass: Object) view viewNamed: 'methods') presenterConnectionPoint allColumns third 
	width: 100;
	getTextBlock: [ :obj | 'test'];
	customDrawBlock: [ :ctx | ctx font isUnderlined: true].`

I have fixes for these issues based on an alternative way of resetting the font; I'll submit a pull request for this.

PR #712, opened 2019-02-25
Fix for #711 (ListView column custom draw issues)

rko281:
- copy the custom draw context's font before any potential modifications
 - reset the canvas to the ListView's font, thus dropping any modifications
 - always reset the context

### DIFF
--- a/Core/Object Arts/Dolphin/MVP/Views/Common Controls/NMLVCUSTOMDRAW.cls
+++ b/Core/Object Arts/Dolphin/MVP/Views/Common Controls/NMLVCUSTOMDRAW.cls
@@ -58,6 +58,12 @@ column: aListViewColumn
 	"Private -"
 	column := aListViewColumn!
 
+font
+	"Answer the character font."
+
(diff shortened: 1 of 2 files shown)


Core/Object Arts/Dolphin/MVP/Views/Common Controls/NMLVCUSTOMDRAW.cls
```diff
@@ -83,9 +89,7 @@ itemHandle
 reset
 	"Reset the canvas to default settings."
 
-	self font
-		isItalic: false;
-		isStruckThrough: false.
+	self canvas font: self view actualFont.
```
blairmcg: Hmm, this doesn't respect the ListView custom draw protocol in the case where the previous column had a custom draw block that changed the font, and the current column does not. It will switch the font back to the default, but it won't return CDRF_NEWFONT to tell the control that the font has been changed. Apparently this doesn't actually matter with the current control implementation (at least on Windows 10, I've not tried it elsewhere), but it might not work on other implementations either now or in the future.

PR #719, opened 2019-03-11
Set https for object arts site - http not working

brunobuzzi:
https://object-arts.com/ - working
http://object-arts.com/ - NOT working


bowerandy:
Both http & https addresses should now be working.

brunobuzzi:
http not working at least for me

Issue #726, opened 2019-04-14
Ability to compare Smalltalk formatted code with Dolphin DifferencesPresenter


brunobuzzi:
Right now you can compare code as plain text. It will be very useful the ability to compare Smalltalk formatted code.
A class like DiffsSmalltalkScintillaStyler should be create in ScintillaStyler hierarchy.

blairmcg:
I think it would probably be better to dispense with the diffs styler altogether and use Scintilla indicators instead. This is the same as the way that brace highlighting and selection matching is done, and would allowing diff highlighting to be independent of the syntax colouring.

brunobuzzi:
For now i have the following:
![image](https://user-images.githubusercontent.com/4719282/56476931-ee17d000-6475-11e9-9fa5-1327a1ec0a06.png)
StDifferencesPresenter as subclass of DifferencesPresenter.
In StDifferencesPresenter there are two **TextStyles**, one is the code style for normal Smalltalk code and the other is ofr Diff Smalltalk code. Both are the same but Diff has `backcolor:` with a value. Also the name of ScintillaTextStyle add Diff:
```
((ScintillaTextStyle name: #illegalDiff)
					id: 100;
					description: 'Illegal characters, e.g. £';
					forecolor: Color red;
					backcolor: Color yellow;
					isBold: true;
					yourself)
```
So different tokens can have different backcolor. The implementation is very simple and a little hackish (the Diff added to scintilla tokens).
DiffSmalltalkMethodStyler (subclass of SmalltalkMethodStyler) re implement:
`colourTokens:startingAt:`
And new method added `applyStyle:from:to:`
```
applyStyle: aSymbol from: start to: end 
	"The receiver apply style named <aSymbol> from token start <start> to token end <end>.
	If the token is contained in <diffs> collection then the style applied is <aSymbol> + <Diff>"
	| style diffSymbol |

	(end - start) < 1 ifTrue: [^self].
	(arrayDepth > 0 or: [aSymbol == #literalArray]) ifTrue: [view setIndicator: self literalArrayIndicatorId from: view stylingPosition length: (end - start)].
	diffSymbol :=  (diffs anySatisfy: [:interval | interval includes: start]) 
				ifTrue: [(aSymbol , 'Diff') asSymbol] 
				ifFalse: [aSymbol].
	style := view styleNamed: diffSymbol.
	view styleNext: (end - start) mask: (style ifNil: [0] ifNotNil: [style id]).
```

[...]

PR #743, opened 2019-04-23
Proposal for #726

blairmcg:
@brunobuzzi, this is syntax colour in the diff browser down with indicators for the diffs instead of requiring a custom styler. It's just a proposal. If you have something better, by all means submit.


brunobuzzi:
@blairmcg give me a few days to post the PR. I will update my local repo to include indicatorDiff branch to check your solution and do the PR to this branch.

PR #746, opened 2019-04-24
Fix for Scintilla Views with forecolor with nil

brunobuzzi:
In ScintillaIndicatorStyle>>stbConvertFrom: if <instance> forecolor is nil --> the STB convertion fails.
This fix make Jade (https://github.com/jgfoster/Jade) to be loaded without any problems with Dolphin latest commit.
Maybe the fix should be in other part but with this all Jade IDE seems to work without problems.


blairmcg:
There is a fix for this issue among a lot of other Scintilla changes in my latest PR.

PR #750, opened 2019-04-25
 Fix decoding of 5 and 6 byte UTF-8 sequences

### LINKED ISSUE #560: UTF-8 decoding should recognise 5 and 6-byte sequences
UTF-8 includes 5 and 6-byte encoded sequences because it was originally designed for the UCS, and earlier standard than Unicode with a wider potential range of code points. These can never represent valid Unicode code point (because the largest Unicode code point is 0x10FFFF, and has a 4-byte encoding), however they are valid in the encoding and so should be decoded, even if the result is the replacement character. This is necessary to read an input stream that contains encoded characters outside the valid Unicode range correctly, otherwise the 5th and 6th trail bytes are interpreted as further invalid characters.

image

For the highlighted cases we should be seeing a single replacement character glyph, not multiple glyphs.

Similarly:
image

blairmcg:
Fixes #560

### DIFF
--- a/Core/Object Arts/Dolphin/Base/Character.cls
+++ b/Core/Object Arts/Dolphin/Base/Character.cls
@@ -180,7 +180,7 @@ codePoint
 		ifTrue: [AnsiCodePoints at: code +1]
 		ifFalse: 
 			["If this is a UTF-8 or UTF-16 surrogate, the code point will be invalid"
-			code bitAnd: 16r1FFFFF]!
+			code bitAnd: 16rFFFFFF]!
 
 codeUnit
 	"Answer the <integer> character code for the character in its native encoding. This can
--- a/Core/Object Arts/Dolphin/Base/PositionableStreamTest.cls
+++ b/Core/Object Arts/Dolphin/Base/PositionableStreamTest.cls
@@ -191,6 +191,10 @@ testNextColon
 			self should: [subject next: 1] raise: Stream endOfStreamSignal.
 			self closeTempStream: subject]!
 
+testNextFiveByteUtf8Encoding
+	"UTF-8 can encode values up to 16rFFFFFFF. In practice these are not required since the maximum Unicode code point is U+10FFFF, which requires only 4 bytes to encode, but in theory they could be encountered, and should be treated as invalid characters (i.e. substituted with the replacement character"
+!
+
 testNextIllegalUtf8CodePositions
 	"Test cases from Kuhn"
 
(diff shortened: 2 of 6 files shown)


mogando668:
@blairmcg  : whatever you've committed as changes for this diff, please undo it. Unicode Consortium explicitly said 5-byte and 6-byte sequences are NOT valid UTF-8 anymore. 

You can find the text verbiage inside latest Unicode spec PDF. 

Attempting to decode 5-byte or 6-byte sequences simply opens up huge cans of security hole worms while removing UTF-8 conformance of it, 

at best bloated up the code base for no reason that causes the function to hallucinate phantom valid sequences out of purely binary data,
at worst created an exploitable vulnerability out of nothing,

all while adding more items to the self testing packages that aren't testing anything useful or meaningful.

blairmcg:
> @blairmcg : whatever you've committed as changes for this diff, please undo it. Unicode Consortium explicitly said 5-byte and 6-byte sequences are NOT valid UTF-8 anymore.

These changes were made six years ago, and I certainly won't be undoing them. Dolphin is an open source system. I work on things that I find interesting, or necessary, or occasionally to help people out who I feel inclined to help. None of these apply. If it bothers you so much, then you could try making a contribution yourself, though before you do that you may want to try and understand the actual intent of the change and its actual realisation rather than "hallucinating" your own false interpretation of it.

> 
> You can find the text verbiage inside latest Unicode spec PDF.
> 

I know. I've read it in detail, as ought to be obvious from the comments in the code and tests.

> Attempting to decode 5-byte or 6-byte sequences simply opens up huge cans of security hole worms while removing UTF-8 conformance of it,
> 
> at best bloated up the code base for no reason that causes the function to hallucinate phantom valid sequences out of purely binary data, at worst created an exploitable vulnerability out of nothing,
> 
> all while adding more items to the self testing packages that aren't testing anything useful or meaningful.

The UTF-8 encoding algorithm invented by Ken Thomson is distinct from whatever limitations Unicode may choose to impose on the encodings it accepts. If you'd bothered to actually look at the change, you may have noticed that the tests you disparage are verifying that 5&6 byte sequences are treated as invalid as part of Unicode strings and have replacement characters substituted following the best practice guidelines outlined in the Unicode standard. Not only is this covered with a specific, though apparently not meaningful or useful, test `Core.Tests.Utf8StringTest>>#testDecodeNextFrom5and6Byte`, but also as part of transcription of the [Markus Kuhn test file](https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt) in the same test class.

PR #759, opened 2019-04-28
Fix methods that don't compile

jgfoster:
DiffsScintillaStyler appears to no longer exist, so removing an extra method.

Move #storeForViewOn:variables: from ToolbarButton to ToolbarBitmapButton (this might not be a complete fix, but the code doesn't compile in the abstract superclass so it has been moved to a concrete subclass with the needed instance variables.

Use accessors instead of direct variable references for variables that no longer exist.

### DIFF
--- a/Core/Contributions/JGFoster/StoreString for Views.pax
+++ b/Core/Contributions/JGFoster/StoreString for Views.pax
@@ -1,4 +1,4 @@
-| package |
+﻿| package |
 package := Package name: 'StoreString for Views'.
 package paxVersion: 1;
 	basicComment: 'Methods in this package are used to generate a string that can be used to recreate a view. The primary use is to allow for diffs when a view is edited (the literal array in a resource_ method is useless for that purpose).'.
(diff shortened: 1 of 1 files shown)


Core/Contributions/JGFoster/StoreString for Views.pax
```diff
@@ -1293,14 +1280,14 @@ storeForViewOn: aStream
 		space;
 		display: #fromMilliseconds:;
 		space.
-	milliseconds storeForViewOn: aStream.
+	self millisecond storeForViewOn: aStream.
```
blairmcg: Time>>millisecond returns the number of milliseconds in the fractional part of the Time’s total number of seconds. To get a precise round trip use asSeconds and fromSeconds:

blairmcg: @jgfoster, I’ll go ahead and merge this code as is if you would prefer to fix it later. I’d guess that having the correct Time instance in a view resource is unlikely to matter. 
I’ll give it a couple more hours then press go, or let me know otherwise.

jgfoster: Either way is fine (no one is complaining about the methods that don't compile). I'll try to get to it today.

PR #779, opened 2019-05-19
Database Connection: Use calculated length for Decimal and Numeric columns

### LINKED ISSUE #778: Database Connection - Incorrect size for Access DECIMAL columns under ODBC3
Following on from #768 and #774 I've found a case where the size of DECIMAL columns is incorrectly reported under ODBC 3. The culprit is Access (again) which is returning 16 as the `SQL_DESC_OCTET_LENGTH` of a DECIMAL column regardless of the declared precision. This can lead to unexpected rounding or errors when fetching results. Example:

```
connection := DBConnection new.
connection dsn: 'AnAccessDSN'.
connection connect.

connection exec: 'create table num_test (num DECIMAL(15,5))'.
(connection exec: 'select * from num_test') describeCols: #(1). 
"#(a DBColAttr(1, num, SQL_NUMERIC, 16))"
connection exec: 'insert into num_test(num) values (-1234567890.12345)'.
(connection query: 'select * from num_test') next. "a DBRow(-1234567890.12292s)"

connection exec: 'drop table num_test'.
connection exec: 'create table num_test (num DECIMAL(20,5))'.
connection exec: 'insert into num_test(num) values (-123456789012345.12345)'.
(connection query: 'select * from num_test') next "ERROR: Index 22 is out of bounds"
```
This didn't occur when using `SQL_COLUMN_LENGTH` under ODBC 2 where Access returned a length of 30 for all DECIMAL columns, which is sufficient for its maximum precision of 28 +2 characters for decimal point and sign. Under ODBC 3 it looks to be necessary to calculate the length for DECIMAL (and NUMERIC) columns as per https://docs.microsoft.com/en-us/sql/odbc/reference/appendixes/transfer-octet-length

Arguably Access is following the standard since the documentation refers to SQL type identifier rather than the related C type (i.e. SQL_DECIMAL rather than SQL_C_CHAR) so using SQL_DESC_OCTET_LENGTH for decimal columns is possibly incorrect, though Access is the only dialect I've encountered which exhibits this problem. 

I'll submit a fix shortly.

rko281:
Reworked #774 to force the use of the calculated length for DECIMAL and NUMERIC columns, and fixed the earlier bug in the previously-unreachable calculation code. For consistency with ODBC documentation the "switch" is now driven by sqlType rather than cType. Also renamed some methods for clarity (hopefully) and removed/renamed redundant ODBC 2 Date/Time types

Tested with the Database Connection 'Northwind' tests, plus the ReStore test suite (with strengthened Decimal tests) across Access, SQLite, MySQL, Postgres and SQL Server.

Closes #778.

### DIFF
--- a/Core/Object Arts/Dolphin/Database/DBColAttr.cls
+++ b/Core/Object Arts/Dolphin/Database/DBColAttr.cls
@@ -7,7 +7,7 @@ Object subclass: #DBColAttr
 	classInstanceVariableNames: ''!
 DBColAttr guid: (GUID fromString: '{87b4c471-026e-11d3-9fd7-00a0cc3e4a32}')!
 DBColAttr addClassConstant: 'CTypesExtraBytes' value: #(0 2 2 2 0 0 0 0 0 0 1 0 1 1 1 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)!
-DBColAttr addClassConstant: 'SQLToCTypes' value: #(-11 -8 -8 -8 -7 -6 -25 -2 -2 -2 1 99 1 1 1 4 5 8 7 8 9 10 11 1 nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil 9 10 11)!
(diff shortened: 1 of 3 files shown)


Core/Object Arts/Dolphin/Database/DBColAttr.cls
```diff
@@ -199,40 +203,39 @@ specialFlagAt: flagMask put: aBoolean
...
-	self type = SQL_BIGINT ifTrue: [^##(20 * 2 "allow for unicode")].
+	sqlType = SQL_BIGINT ifTrue: [^##(20 * 2 "allow for unicode")].
 
 	"SQL_DECIMAL, SQL_NUMERIC
 	The number of bytes required to hold the character representation of this data if the character set is ANSI, and twice this number if the character set is UNICODE. This is the maximum number of digits plus two, because the data is returned as a character string and characters are needed for the digits, a sign, and a decimal point. For example, the transfer length of a column defined as NUMERIC(10,3) is 12."
-	(self type = SQL_DECIMAL or: [self type = SQL_NUMERIC]) ifTrue: [^self length: (self precision + 2) * 2 "allow for unicode"].
+	(sqlType = SQL_DECIMAL or: [sqlType = SQL_NUMERIC]) ifTrue: [^(self precision + 2) * 2 "allow for unicode"].
```
blairmcg: I think it would be find to just simplify this away and return the largest possible size, then you wouldn't need the switch :-). The maximum size is relatively small and the data will need to be transferred as a string and parsed anyway, so the benefit in being precise is small. It's not worth spending more time on though, and good to bottom it out. 
It makes me think it would be useful to allocate some slop around byte objects in the debug VM and fill it with a pattern that could be checked for overwrites on GC. Getting the size of these buffers wrong has unpleasant effects as all this case has demonstrated.

PR #783, opened 2019-05-28
Fix test to run in non-UTC zone

jgfoster:
Depending on when I run the tests, I often get an error in DateTest>>testFromSeconds. I believe that the difference is that I'm not in the UTC time zone.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/DateTest.cls
+++ b/Core/Object Arts/Dolphin/Base/DateTest.cls
@@ -43,7 +43,7 @@ testFromSeconds
 	today := Date today.
 	date := Date fromSeconds: today asSeconds.
 	self assert: today equals: date.
-	dtNow := DateAndTime now.
+	dtNow := DateAndTime nowUTC.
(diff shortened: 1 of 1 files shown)


Core/Object Arts/Dolphin/Base/DateTest.cls
```diff
@@ -43,7 +43,7 @@ testFromSeconds
 	today := Date today.
 	date := Date fromSeconds: today asSeconds.
 	self assert: today equals: date.
-	dtNow := DateAndTime now.
+	dtNow := DateAndTime nowUTC.
```
blairmcg: I had to think for a bit to understand why this test was failing and whether it was right to change it this way, before concluding that it is. It fails because:
- DateAndTime>>asSeconds always returns a UTC relative number of seconds since the epoch
- the Date class is not timezone aware
- DateAndTime>>dayOfMonth will return the day of the month in local time. This could be a different day than that calculated by basing off the epoch in UTC seconds.
Since we aren't trying to test DateAndTime>>asSeconds here, this is the right thing to do.

PR #790, opened 2019-06-07
Fix for DateAndTime class>>stbConvertFrom: 

jgfoster:
My stored instances of DateAndTime are not converted properly by the existing code. This fixes it for me. I've left the original code and tests, but made the conversion conditional on finding the expected data. I suspect that this is not the ideal situation, and welcome guidance on why the existing code is broken for me, as well as any suggestions on how to improve the proposed code (implementation or style).

### DIFF
--- a/Core/Object Arts/Dolphin/Base/DateAndTime.cls
+++ b/Core/Object Arts/Dolphin/Base/DateAndTime.cls
@@ -656,10 +656,21 @@ stbConvertFrom: anSTBClassFormat
 
 	^
 	[:data |
-	| totalSeconds offset |
-	offset := data second.
-	totalSeconds := data first - 59958230400 - offset asSeconds.
-	self utcSeconds: totalSeconds offset: offset asSeconds]!
+		(data at: 2) class == Duration ifTrue: [		"Original code satisfies one set of tests"
+			| totalSeconds offset |
+			offset := data second.
+			totalSeconds := data first - 59958230400 - offset asSeconds.
+			self utcSeconds: totalSeconds offset: offset asSeconds.
+		] ifFalse: [							"This satisfied another format found in the wild"
+			| days seconds offset totalSeconds |
+			days := data at: 1.		"Days since Sunday, December 31, 0000"
+			days := days - 693961.	"Days since December 31, 1900"
+			seconds := data at: 2.	"Likely to be a Fraction"
+			offset := data at: 3.
+			totalSeconds := days * 86400 + seconds - offset.		"86400 == (24 * 60 * 60)"
+			self utcSeconds: totalSeconds offset: offset.
+		]
+	]!
 
 stbVersion
 	"Answer the current binary filer version number for instances of the receiver."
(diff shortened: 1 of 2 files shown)


Core/Object Arts/Dolphin/Base/DateAndTimeTest.cls
```diff
@@ -486,6 +486,36 @@ testStbConvertFrom
...
+	dat binaryStoreBytes.
+	 "
+
+	| bytes dat string |
+	bytes := #[33 83 84 66 32 52 32 6 3 11 0 68 97 116 101 65 110 100 84 
+		105 109 101 131 127 22 0 6 2 8 0 70 114 97 99 116 105 111 110 183 
+		217 116 4 209 7 0 0 33 59 255 255].	"Obtained from the code above"
+	dat := DateAndTime fromBinaryStoreBytes: bytes.
+	string := dat printString.
+	self assert: string = ' 2019-06-07T10:23:03.387-07:00'.	"What is with the leading space?"
```
blairmcg: ANSI NCITS 319-1998, <DateAndTime> protocol definition, section 5.8.1.24, Message refinement: printString
```
...
The result will be formatted as
 -YYYY-MM-DDThh:mm:ss.s+ZZ:zz.z where
 -     is the <Character> $- if the year is less than 0 otherwise it is the <Character> that is 
        returned from the message #space sent to the global Character,
...
```
It means that the result is always of the same width (assuming the same precision is used for fractional seconds), but I find it a strange choice too. You can omit it by dropping down a level to #printOn:withLeadingSpace:, or use `#printStringFormat:` to have complete control over the format.

Core/Object Arts/Dolphin/Base/DateAndTime.cls
```diff
@@ -656,10 +656,21 @@ stbConvertFrom: anSTBClassFormat
...
+		(data at: 2) class == Duration ifTrue: [		"Original code satisfies one set of tests"
+			| totalSeconds offset |
+			offset := data second.
+			totalSeconds := data first - 59958230400 - offset asSeconds.
+			self utcSeconds: totalSeconds offset: offset asSeconds.
+		] ifFalse: [							"This satisfied another format found in the wild"
+			| days seconds offset totalSeconds |
+			days := data at: 1.		"Days since Sunday, December 31, 0000"
+			days := days - 693961.	"Days since December 31, 1900"
+			seconds := data at: 2.	"Likely to be a Fraction"
```
blairmcg: These instances are not from Ian's original definition. They are from the window between PR's #741 and #748 - hence why an instance created with code from 24th April will be of this form, as that is in the window between the PRs.
They should not require any data conversion because the inst var values should be correct as is. Consequently the code can be simplified to:

```
	^
	[:data |
	| iv2 |
	iv2 := data second.
	iv2 class == Duration
		ifTrue: 
			[| totalSeconds |
			"Convert from IDB format DateAndTime"
			totalSeconds := data first - 59958230400 - iv2 asSeconds.
			self utcSeconds: totalSeconds offset: iv2 asSeconds]
		ifFalse: 
			["The object is already of the right format, but was saved after the addition of DateAndTime to the base but before the STB version upgrade for compatibility with the original IDB DateAndTimes was added."
			self basicNew
				setRataDie: data first
				seconds: iv2
				offset: data third]]
```

PR #793, opened 2019-06-14
Fix #792 Date>>print:on:format:alphabet:formatters: not escaping $' c…

### LINKED ISSUE #792: Date>>print:on:format:alphabet:formatters:  not escaping $' character
**Description**
When `Date>>print:on:format:alphabet:formatters:` has a formatString like: 'dddd, d ''de'' MMMM ''de'' yyyy' ( a WindowsLocale('es-AR') longDateFormat). I was expecting it should escape the $' characters.

**Steps to reproduce the behavior:**
Print the result of the following:
`stream := String writeStream: 32.
(Locale lcid: 11274) printDate: Date today on: stream format: 'dddd, d ''de'' MMMM ''de'' yyyy'.
stream contents.`
and you get:
`'viernes, 14 ''14e'' junio ''14e'' 2019'`

**Expected behavior**
Should get:
`'viernes, 14 de junio de 2019'`

**Screenshots**
image

**OS information:**
 - OS: Windows 10

**Additional information**
WindowsLocale('es-AR') lcid is: 11274.

msobral:
…haracter.

This change should fix the bug.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/Locale.cls
+++ b/Core/Object Arts/Dolphin/Base/Locale.cls
@@ -435,16 +435,12 @@ print: anObject on: aWriteStream format: formatString alphabet: formatChars form
 	formatStream := formatString readStream.
 	ch := formatStream nextAvailable.
 	[ch isNil] whileFalse: 
-		[ch == $'
-			ifTrue: 
-				[aWriteStream nextPutAll: (formatStream upTo: $').
-				ch := formatStream nextAvailable].
-		(formatChars includes: ch)
+		[(formatChars includes: ch)
 			ifTrue: 
 				[| start |
 				tokenStream reset.
 				start := ch.
-					
+				
 				[tokenStream nextPut: ch.
 				ch := formatStream nextAvailable.
 				ch = start] whileTrue.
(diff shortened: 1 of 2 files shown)


Core/Object Arts/Dolphin/Base/Locale.cls
```diff
@@ -429,29 +429,34 @@ pmDesignator
...
-			[(formatChars includes: ch)
-				ifTrue: 
-					[| start |
-					tokenStream reset.
-					start := ch.
+		[ch == $'
+			ifTrue: 
+				[aWriteStream nextPutAll: (formatStream upTo: $').
+				ch := formatStream nextAvailable].
+		(formatChars includes: ch)
```
blairmcg: There is a bug here. This change does need a regression test. I would suggest modifying the existing `LocaleTest>>testPrintDateTimeOnFormat` as follows:

```
testPrintDateTimeOnFormat
	| subject now expected |
	now := DateAndTime
				year: 2019
				day: 97
				hour: 18
				minute: 26
				second: 21 + (827 / 1000).
	subject := self defaultSubject.
	expected := '6:26 ', subject pmDesignator, ' (and 21.827 seconds) on Sunday, 7 of April 2019 somewhere'.
	{now. now asParameter. TimeStamp fromDateAndTime: now} do: 
			[:each |
			| actual |
			actual := String streamContents: 
							[:s |
							subject
								printDateTime: each
								on: s
								format: 'h:mm tt (an\d ssffffff \secon\d\s) on dddd, d ''of'' MMMM yyyy ''somewhere'''].
			self assert: actual equals: expected]
```

By having one of the quoted parts of the format string at the end, it will show up the bug.

msobral:
LocaleTest>>testPrintDateTimeOnFormat modified as suggested.
This test passes with this other solution thanks for the review.

PR #800, opened 2019-08-01
7.0

Fabro-f:


blairmcg:
I'm not sure why you are trying to merge from 7.0 into master, but its not a good idea. 7.0 is an old version from which master has diverged.

PR #813, opened 2019-08-23
STS folder removal causes error in HelpImagesCreationTest>>testSystemFolder

brickviking:
In the process of generating help images for future help docs, I ran across an error while generating a PNG of the STS folder contents. As the Source Tracking Tools have generally been removed, so has the folder containing them, hence the problem. I've commented out the STS folder image PNG export, which should fix that bug at least. I've also changed the background colour of the System Folder window to be more like the modern Dolphin 7 colour.

More bugs to come.

Incidentally, how on earth can I simply include JUST the files I want in the pull request, not everything else I've done previously? For the life of me I can't figure out how to get github to do that.


blairmcg:
> Incidentally, how on earth can I simply include JUST the files I want in the pull request, not everything else I've done previously? For the life of me I can't figure out how to get github to do that.

Here is what I do. There are other approaches, but this works for me:

- Follow the topic branch model and create a new branch off master for each work item. When its merged, delete it. Then 
- Never do `git pull` as this will create merge commits. I don't like having these in the history because they make investigating the history much more difficult, so if you PR contains them I will `squash` it to remove them when I commit it, losing the individual commits but maintaining the single path of changes. If there are no merge commits then I will rebase it when I apply it, keeping your individual commit history. I would only do an actual merge in very rare circumstances where it was desirable to retain the original commit history precisely.
- If you need to update your topic branch from the upstream separately `git fetch` it and rebase onto the new upstream master. 
- Use `git rebase -i` to understand what is going to happen when you rebase, and to edit the local commit history to clean it up if you want to do that.
- If you make a mistake you can amend the last commit easily, but for ones before that you can add a fixup commit with `git commit --fixup <sha>` and then rebase to collapse the fixup into the original change you want to modify, i.e. do a `git rebase -i` and it will be automatically set up as a fixup for the correct commit, which you can cross check in the editor before proceeding.
- Update your clone from the main repo by fetching from the upstream and pushing directly to origin/master rather than merging. I usually add a separate 'upstream' remote to my local repo configuration to make it easier to fetch from the upstream.

Core/Object Arts/Dolphin/IDE/Base/HelpImagesCreationTest.cls
```diff
@@ -321,20 +321,20 @@ testSystemFolder
...
-	stsTools := folderPresenter list detect: [:each | each description = 'Source Tracking Tools'].
+	"stsTools := folderPresenter list detect: [:each | each description = 'Source Tracking Tools']."
 	samples := folderPresenter list detect: [:each | each description = 'Sample Applications'].
 	additionalTools := folderPresenter list detect: [:each | each description = 'Additional Tools'].
 	"Top level icon images"
 	self saveIconImagesInFolder: folderPresenter.
 	"Source Tracking Tools images"
-	folderPresenter expand: stsTools.
-	self saveIconImagesInFolder: folderPresenter.
+	"folderPresenter expand: stsTools.
```
blairmcg: I suggest deleting this section entirely. If the STS tools are every brought back to life, they'd be a separate thing.

Core/Object Arts/Dolphin/IDE/Base/HelpImagesCreationTest.cls
```diff
@@ -321,20 +321,20 @@ testSystemFolder
 	(folderPresenter view)
 		lvmSetBkImage: LVBKIMAGEW new;
 		backcolor: (Color 
-					red: 255
-					green: 255
-					blue: 238).
+					red: 41
```
blairmcg: Use `(Color named: #smalltalkSystem)` rather than hardcoding. I know it was done that way before, but by using the symbolic colour it won't need to be updated again if this is changed.

brickviking:
Right. That's the last of that done so far. I found something else to do with STS as well, I'll deal with that shortly. Thank you for the comments regarding git, I'll see how well I can put them into practice.

PR #822, opened 2019-09-11
Upgrade build to VS2019

### LINKED ISSUE #724: Target Visual Studio 2019 and latest SDK
Visual Studio 2019 is out, and it is good te have compatibility with it.
I've tested it by first deinstalling VS2017 and all SDKs.
Then installing VS2019 and targeting the latest SDK.
Only one line needed to be changed:
Str.h@57: Added cast to supported << overload.
	return stream << (char*) static_cast<LPCWSTR>(A2W(str.c_str()));
Added the (char*) to make it work, but am not sure about char* vs wchar_t*.
All regression tests ran fine after that.

blairmcg:
Fixes #724


FunctionPoint:
Great, tnx!

Issue #535, opened 2018-06-18
Use C99 fixed-width C integer types


blairmcg:
_From @jgfoster on March 28, 2018 19:16_

The C99 standard includes definitions of several new integer types to enhance the portability of programs (per [Wikipedia](https://en.wikipedia.org/wiki/C_data_types#Fixed-width_integer_types)). I'm looking for some simple tasks where students could make a contribution. This seemed like a reasonable one so I'll create an issue to track it.

_Copied from original issue: dolphinsmalltalk/DolphinVM#163_

jgfoster:
I'm confused by the move of this enhancement from the VM to the image. Wouldn't this require changes to the VM (or C code?).

blairmcg:
The repos have been merged together, see #530. I did leave the PR open for several days for comments. If you have any work in progress on the VM, don't worry however, it can still be submitted via the DolphinVM repo if you want, and it can be merged in easily afterwards. Just let me know what you need to do and I'll try to help.

jgfoster:
Excellent! Sorry I wasn't keeping up on the latest changes.

blairmcg:
This should be completely done now.

Issue #531, opened 2018-06-18
Enhance the VM to 64bit


blairmcg:
_From @objectarts on January 28, 2016 17:56_

_Copied from original issue: dolphinsmalltalk/DolphinVM#22_

blairmcg:
_From @jgfoster on January 29, 2016 14:5_

@objectarts, I'm interested in this task and it gives me the opportunity to raise a question about coding style and names. 

For me there is a non-trivial cognitive load to translating all the integer types into signs and sizes. How big is `int`? Is it the same in 32-bit and in 64-bit? How is it different from `long` (see compiler.cpp:723)? What is `DWORD`? I know the size of BYTE but is it signed or unsigned? A certain number of the runtime errors (#18) are related to conversion to smaller types and I can't help but imagine that it would be useful to have these things more obvious.

In the past I've been involved in projects where typdefs were created for all integer types along the lines of the following: `sintNN` and `uintNN` where `NN` is 8, 16, 32, 64, or 128. This allows one to tell at a glance whether an item is signed or unsigned and its size.

I've assumed that most people with experience on a project get used to the existing names and would be reluctant to see things change, but I still think that it requires some cognitive load (and "[brainpower is by far our scarcest resource](https://www.cs.utexas.edu/~EWD/transcriptions/EWD03xx/EWD340.html)"). 

This enhancement request has given me courage to at least ask your thoughts on such a change. Should I immerse myself in the environment and get used to the existing names (and create a cheat-sheet to help carry the cognitive load), or would you be receptive to considering some rather large-scale (but otherwise trivial) edits? I expect the answer to be that I should follow the existing conventions, but I thought I'd ask...

blairmcg:
@jgfoster, this is a huge work item hiding behind a very short description. The VM contains large amounts of x86 assembly code which needs wholesale replacement one way or another.

I did make a start on it a few years ago, and you can see some of the residue of this in the FFI support in the image, e.g. the ExternalField type INT_PTRField which varies in size, and this and its unsigned equivalent, are used in various ExternalStructure definitions that contain pointers or integer values that vary in length on Win64 vs Win32. The VM passes in the actual value through the well known VMConstants pool, and it was set up so that the structures would recompile to match the actual pointer size (not sure if that still works). All the work I did on the VM, which would have started by removing 32/64 bit portability issues like the kind of issues you mention, has long since been lost.

[...]

PR #854, opened 2019-12-03
Use fixed width integer types in Compiler

blairmcg:
Generally replace BYTE, DWORD, etc, with corresponding C standard types uint8_t, uint32_t, etc.
Also, for future 64-bit compatibility, replace unsigned and int in most cases with an integer type that is the same as the machine word size. For example size_t is used for variables that can be used as indexes and memory sizes.
- Contributes to #535
- Contributes to #531
- Also address many code analysis issues from VS19 in-IDE parser
  - e.g. Use of enum class in preference to unscoped enums
- Define specific integer types (using enum class) for IPs and text positions to help identify
errors

### DIFF
--- a/Core/DolphinVM/Compiler/compiler.cpp
+++ b/Core/DolphinVM/Compiler/compiler.cpp
@@ -71,47 +71,41 @@ LPUTF8 s_restrictedSelectors[] =
 ///////////////////////
 
 
-Compiler::LibCallType Compiler::callTypes[DolphinX::NumCallConventions] = 
+Compiler::LibCallType Compiler::callTypes[4] =
 {
-	(LPUTF8)"stdcall:", DolphinX::ExtCallStdCall,
-	(LPUTF8)"cdecl:", DolphinX::ExtCallCDecl,
-	(LPUTF8)"thiscall:", DolphinX::ExtCallThisCall,
-	(LPUTF8)"fastcall:", DolphinX::ExtCallFastCall
+	(LPUTF8)"stdcall:", DolphinX::ExtCallDeclSpec::StdCall,
+	(LPUTF8)"cdecl:", DolphinX::ExtCallDeclSpec::CDecl,
+	(LPUTF8)"thiscall:", DolphinX::ExtCallDeclSpec::ThisCall,
+	(LPUTF8)"fastcall:", DolphinX::ExtCallDeclSpec::FastCall
 };
 
 ///////////////////////
 
 Compiler::Compiler() :
-		m_allScopes(NULL),
-		m_bytecodes(NULL),
 		m_class(0),
-		m_codePointer(0),
-		m_compiledMethodClass(NULL),
+		m_codePointer((ip_t)0),
+		m_compiledMethodClass(nullptr),
 		m_compilerObject(0),
 		m_context(0),
-		m_flags(Default),
-		m_instVars(NULL),
+		m_flags(CompilerFlags::Default),
 		m_instVarsInitialized(false),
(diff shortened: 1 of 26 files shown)


Core/DolphinVM/Compiler/compiler.cpp
```diff
@@ -3830,16 +3814,16 @@ TempVarDecl* Compiler::DeclareTemp(const Str& strName, const TEXTRANGE& range)
...
 	m_pCurrentScope->SetTextStop(textStop);
-	m_pCurrentScope = m_pCurrentScope->GetOuter();
+	m_pCurrentScope = m_pCurrentScope->Outer;
 }
 
-void Compiler::PushNewScope(int textStart, bool bOptimized)
+void Compiler::PushNewScope(textpos_t textStart, bool bOptimized)
 {
-	int start = textStart < 0 ? ThisTokenRange().m_start : textStart;
+	textpos_t start = textStart < (textpos_t)0 ? ThisTokenRange.m_start : textStart;
```
blairmcg: textpos_t::start

Core/DolphinVM/Compiler/compiler.cpp
```diff
@@ -3853,70 +3837,70 @@ void Compiler::PushNewScope(int textStart, bool bOptimized)
...
 }
 
-void Compiler::PushOptimizedScope(int textStart)
+void Compiler::PushOptimizedScope(textpos_t textStart)
 {
 	PushNewScope(textStart, true);
 }
 
-inline BYTE MakeOuterTempRef(int blockDepth, int index)
+inline BYTE MakeOuterTempRef(size_t blockDepth, int index)
```
blairmcg: Unused, remove

Core/DolphinVM/ist.h
```diff
@@ -75,15 +75,19 @@
...
+typedef _Return_type_success_(return >= 0) int32_t NTSTATUS;
+#define NTSTATUS_DEFINED
+#define _NTDEF_
+
 #include "Environ.h"
 
-typedef signed char		SBYTE;
-typedef short			SWORD;
-typedef long			SDWORD;
+typedef int8_t		SBYTE;
```
blairmcg: Eventually these will be removed, but at present they are still used elsewhere in the VM

Core/DolphinVM/Compiler/lexer.cpp
```diff
@@ -606,66 +606,66 @@ Lexer::TokenType Lexer::NextToken()
...
 		else if (isAnsiBinaryChar(ch))
 		{
-			m_tokenType = Binary;
+			m_tokenType = TokenType::Binary;
 		}
 		else
 		{
-			int pos = CharPosition();
+			textpos_t pos = CharPosition;
 			int cp = ReadUtf8(ch);
```
blairmcg: int32_t

Core/DolphinVM/Compiler/optimizer.cpp
```diff
@@ -978,37 +980,38 @@ int Compiler::CombinePairs1()
...
-		if (next >= GetCodeSize())
+		const ip_t next = i+len1;
+		if (next > LastIp)
 			break;					// bytecode1 is last instruction
 		BYTECODE& bytecode2=m_bytecodes[next];
 		
-		_ASSERTE(bytecode1.isOpCode() && bytecode2.isOpCode());
+		_ASSERTE(bytecode1.IsOpCode && bytecode2.IsOpCode);
 
 		BYTE byte1 = bytecode1.byte;
```
blairmcg: uint8_t

PR #855, opened 2020-01-23
Debugger improvements

daniels220:
Two significant improvements and one minor bugfix to the Debugger. The "Step Into Block" command needs an icon, but without the resources used to create the others I couldn't make one that would fit in. If there are source files (SVG, perhaps?) I could make one and add that to the PR.


Core/Object Arts/Dolphin/IDE/Base/Debugger.cls
```diff
@@ -437,7 +437,7 @@ cursorIPRange
 	self makeDebugFrame: frame.	"first make sure it is a debug frame."
 	pos := sourcePresenter view caretPosition.
 	map := frame textMap.
-	i := map findFirst: [:a | a value includes: pos].
+	i := map findFirst: [:a | a value stop > pos].
```
blairmcg: Can you explain what the effect of this is please, with some examples? I’d like to understand why this is necessary. I’m also concerned about regressing other cases, given that there are no existing tests.

blairmcg: I read your commit comment, which does explain. I’d want to put this behind an option as I am used to the current behaviour and as it has been as it is for many years, I am probably not alone.

daniels220: I'm honestly curious—have you tried the changed code and found a case where you expect/prefer the current behavior? I used the stock behavior for a while and was always frustrated because I honestly could not figure out how it actually decided where to stop. Since the change I've never been frustrated by the new behavior—and, at a glance, I don't think it will ever proceed *further* than the old behavior, so at worst, it means a couple extra `Step Over`s.

Or, putting it behind an option would work, but I'm not sure what to call the option—how to describe what it does in less than the paragraph it took in the commit message. Honestly it seems like a case where the psychological cost of having the option exceeds the value of it—but I might think otherwise if I had a better sense how you use it, or if you have an idea for how to clearly explain it.

rko281:
My [MethodSpies](https://github.com/rko281/MethodSpies) package has a step into block function which I'll remove if/when this goes through. 

You're welcome to use the [icon](https://www.dropbox.com/s/cta688408mokmnx/stepintoblock.ico?dl=1) I designed for this which matches the current Debugger set:

Core/Object Arts/Dolphin/IDE/Base/Debugger.cls
blairmcg: I'm not particularly well placed to judge whether it is or is not better because it is solving a problem I didn't know I had. I'm used to the existing behaviour and it does what I now expect. Whether it has always done what I expect, or whether I'm just conditioned to the way it works, I'm not sure, although I think I would have changed it in the early days if it was surprising me. And that is the crux of the matter - it's been the same since the mid 1990s, so I don't want to change the behaviour without being convinced that the benefit outweighs the cost of those used to the existing behaviour having to relearn a new behaviour.

This isn't really the place for a long discussion of the "what" anyway, as code reviews are for the "how". Any change should really be associated with an issue that precedes it and which would provide a better place for discussion. This time I'm happy to accept the change, but I prefer seeing enhancements proposed in an issue before a PR is sent.

Once this PR is merged, I will put it behind an option, and leave the default as is for now. If in the light of experience there is a general consensus of opinion that your revised behaviour is a better default, then we can make that the default instead.

PR #857, opened 2020-01-23
Move BoundsError under NotFoundError

daniels220:
It shares the same notion of "receiver", and the "is-a" relationship seems justified--ask for element 11 of a size-10 Array, well, it isn't found, same as if you'd asked for key 11 in a Dictionary that doesn't contain it.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/BoundsError.cls
+++ b/Core/Object Arts/Dolphin/Base/BoundsError.cls
@@ -1,7 +1,7 @@
 ﻿"Filed out from Dolphin Smalltalk 7"!
 
-Error subclass: #BoundsError
-	instanceVariableNames: 'receiver'
+NotFoundError subclass: #BoundsError
+	instanceVariableNames: ''
 	classVariableNames: ''
 	poolDictionaries: ''
 	classInstanceVariableNames: ''!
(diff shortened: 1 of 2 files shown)


blairmcg:
Nobody has commented, but my view is that this change should not be made. The reasons are:

1. This is effectively an API change. Although it is not common to catch `BoundsError` (except in tests), it isn't that uncommon to catch `NotFoundError`. Now if the original author hadn't expected catching `NotFoundError` to also catch `BoundsError`, then their code may be broken and in a way that may be quite hidden. 
1. There isn't really a lot of advantage, as there are other convenient ways to catch more than one type of exception with the same handler, e.g. 
```smalltalk
[collection at: key] on: NotFoundError, BoundsError do: [:ex | self log: 'Key not found']
```

Although I don't think there is a lot of value, a way that this change could be made without causing an implicit API change. If a new abstract superclass were introduced for both `NotFoundError` and `BoundsError`, then it could be caught to achieve the desired effect in new code, with a much lower risk of altering the behaviour of old code.

daniels220:
Yeah, fair enough. It made sense to me at the time, and I still think in a vacuum it would be the right thing to do, but I can see how it's not worth the breaking change...

PR #862, opened 2020-02-15
Add Dolphin Sockets Tests to the regression tests

daniels220:
There's not much there, but sure, why not run it. Had to remove a couple assertions about host names--it's hard to come up with a host name that has a stable IP address and reverse-DNS record...

### DIFF
--- a/Core/Object Arts/Dolphin/Sockets/InternetAddressTest.cls
+++ b/Core/Object Arts/Dolphin/Sockets/InternetAddressTest.cls
@@ -1,4 +1,4 @@
-"Filed out from Dolphin Smalltalk 7"!
+﻿"Filed out from Dolphin Smalltalk 7"!
 
 DolphinTest subclass: #InternetAddressTest
 	instanceVariableNames: ''
@@ -17,19 +17,17 @@ testAllForHost
 	"Microsoft should have more than one IP associated with microsoft.com"
 	addresses := InternetAddress allForHost: 'microsoft.com'.
 	self assert: addresses size > 1.
-	self assert: (addresses anySatisfy: [:each | each ipAddress beginsWith: #[191 239]]).
 
 	"OA central"
 	addresses := InternetAddress allForHost: 'object-arts.com'.
-	self assert: addresses size = 1.
-	self assert: (addresses first host endsWith: 'cable.virginm.net')!
+	self assert: addresses size = 1!
 
 testHost
 	| name ip |
 	self isOnline ifFalse: [^self].
 	name := InternetAddress host: 'uk2.net'.
 	ip := InternetAddress ipAddress: name ipAddress.
-	self assert: ip host = name host!
+	self assert: (ip host endsWith: name host)!
 
 testIsIPString
 	self deny: (InternetAddress isIPString: '').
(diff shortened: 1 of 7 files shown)


daniels220:
Huh, I see that `InternetAddressTest>>testLocalHost` failed in the CI build. It works on my machine, which I'm guessing comes down to a difference in configuration. The test should obviously not be that brittle, but I don't even know *how* exactly it's failing in CI. Any thoughts?

blairmcg:
How about google-public-dns-a.google.com?

Regarding the failing test, I would start by adding some descriptions to the assertions in the test, and also replacing any generic assert: expressions with assert:equals:, etc, where appropriate. You might also want to consider setting up an appveyor build on your fork to make it easier to run a bunch of test commits through builds. This is relatively straightforward as the build is described in a checked in yaml file.

Core/Object Arts/Dolphin/Sockets/InternetAddressTest.cls
```diff
@@ -17,19 +17,17 @@ testAllForHost
 	"Microsoft should have more than one IP associated with microsoft.com"
```
blairmcg: I suggest replacing with: 
```smalltalk
	"dns.google has two fixed and well-known IP addresses"
	addresses := InternetAddress allForHost: 'dns.google'.
	self assert: addresses size equals: 2.
	self assert: (addresses collect: [:each | each ipAddress]) asSet
		equals: #(#[8 8 4 4] #[8 8 8 8]) asSet.
```

Core/Object Arts/Dolphin/Sockets/InternetAddressTest.cls
```diff
@@ -17,19 +17,17 @@ testAllForHost
...
 testHost
 	| name ip |
 	self isOnline ifFalse: [^self].
 	name := InternetAddress host: 'uk2.net'.
 	ip := InternetAddress ipAddress: name ipAddress.
-	self assert: ip host = name host!
+	self assert: (ip host endsWith: name host)!
 
 testIsIPString
 	self deny: (InternetAddress isIPString: '').
```
blairmcg: Replacing testLocalHost with something like the below will make it clear which of the assertions is failing:
```smalltalk
testLocalHost
	| local computerName |
	computerName := SessionManager current computerName asLowercase.
	local := InternetAddress localHost.
	self assert: local host asLowercase equals: computerName.
	self assert: ((InternetAddress ipAddress: local ipAddress) host subStrings: $.) first asLowercase
		equals: computerName
```

daniels220:
Good ideas, thanks—trying them out, we'll see how it goes with `testLocalHost`. Is there a problem with just using the automatic AppVeyor build here? If it costs you money or bothers you with emails or something I can set up my own, but otherwise that seems like what it's for...

blairmcg:
No problem at all. The only reason I suggested it was because finding there are errors after raising a PR is a little late. By the time I raise a PR I prefer to know that I'm done apart from responding to code review comments, but please work in whatever way you prefer.

BTW: I don't think it would be unreasonable to add a 'self skipIfCiBuild' to the failing test and raise an issue to track it. It still represents a useful improvement.

blairmcg:
As this has been waiting here for a month and a half, I'm going to merge it even though that will break the build. I'll then add a skip for the test that doesn't run on the CI machine.

daniels220:
Sorry for letting this slip through the cracks—I more or less meant to do that (add a skip) myself, and just kept not getting around to it. Thanks for taking care of it.

PR #890, opened 2020-05-07
Upgrade to Scintilla 4.3.3

blairmcg:


fxgallego:
@blairmcg Is there an option to autoclose a predefined set of characters? It is something that should be done client side? I got used to mainstream editors and when I press " ' [ I'm always waiting for the autoclose char. Also got used to CTRL+D to duplicate lines which is correctly defined in Scintilla keybindings but in workspaces it's used for "Display it". Maybe a better default would be ctrl+alt+D

blairmcg:
As you suspect that would be a something done at the level of the Scintilla host application, not by Scintilla itself. Scintilla provides the toolkit, but generally not high-level editing behaviours.  As an example see `ScintillaView>>braceHighlight`. The Scintilla control provides both a way to search for matching brackets, and a way to highlight them, but it doesn't tie it altogether into a brace matching feature, even as a default. There is a deliberate architectural layering that the Scite example editor helps to delineate.
It would be entirely possible to do that in the Smalltalk integration layer - probably harder to decide exactly how you want it to work, than to actually do it.
As for the Ctrl+D clash, yes that is unfortunate. The Dolphin workspace shortcuts were decided long before Scintilla was integrated. I believe Ctrl+D to duplicate lines was probably following the Visual Studio convention, which didn't exist when we started on Dolphin either. I don't remember whether the early VC++ editions we used had that key binding, but if so we either ignored it or didn't notice. It does seem rather late to change such longstanding choice now. What is really needed is better customisation capability so that you could both change it to suit your preference, and also have that persisted in a convenient way that doesn't get lost and which gets picked up in any new image. At present most of the workspace key bindings are not even easily configurable because they are stored in the menu definitions in the view resources.

fxgallego:
Yes it's a bit tricky to autoclose but we can adhere to Visual Studio Code behavior. I didn't notice it strange.
For the Ctrl + D my suggestion was changing the duplicate command to Ctrl + Alt +D instead of changing #displayIt

PR #894, opened 2020-05-12
Prevent error snapshotting image with visible Scintilla markers

### LINKED ISSUE #893: Image snapshot fails when Scintilla markers are visible
Attempting to snapshot the image whilst there is an active Scintilla marker gives an error; the image is not saved.

**To Reproduce**
1. Start image, open a browser
2. Enter any nonsense code in a method definition to trigger a Scintilla warning/error marker (e.g. `self nonsenseMessage`)
3. Attempt to snapshot the image.
4. MNU: WriteStream does not understand #add:

Error is due to use of `add:` in `ScintillaView>>#getRawAnnotations` introduced in 6ac4a11.

rko281:
Change `#add:` to `#nextPut:`. Fixes #893.


blairmcg:
I should've added a test for this when I changed it, which would have picked up the break. I'll do that.

PR #913, opened 2020-05-28
Fix #898

### LINKED ISSUE #898: Opening a MessageBox to confirm a ListView selection change causes ListView to steal the event loop until the mouse is moved
Open something with a MethodBrowser—a class browser, method explorer, etc—with more than one method showing. Click on a method, make a change but don't accept it. Then carefully click on another method, without moving the mouse after you click. Answer "no" when asked if you want to save changes. The list selection will change, but the method workspace at the bottom doesn't update—until you move the mouse, even by a pixel, at which point it does.

I believe the process is getting stuck performing `aMouseEvent defaultWindowProcessing` in `ListView>>onButtonPressed:`—the ListView takes over the event loop waiting for a mouse-up event (while e.g. watching for movement to initiate a drag), but the mouse-up event has already been processed by the MessageBox. When the mouse is moved, I guess it notices that okay, I haven't seen a mouse up, but the mouse button *isn't down* so I better let go of the event loop. ...though this doesn't *exactly* fit, because if you move the mouse *while the MessageBox is still open*, the problem doesn't occur...

It's not obvious to me what the Right Way To Do This is, but opening a MessageBox to confirm a list selection change seems like something that other Win32 programmers would want to do, even outside of Dolphin, so I would *think* there must be a way to make it work...

blairmcg:
@daniels220, you may want to try this out to see if it fixes your scenario. The eventual fix is quite simple and is similar to what was discussed. It works for the new tests that reproduced the issues of blocking on the mouse down, and the duplicate selection changing events.

A test for marquee selection is still pending, but this will still fire two selection changing events - one for initial loss of any selection, and the 2nd for the eventual new selections.

### DIFF
--- a/Core/Contributions/Camp Smalltalk/SUnit/OA SUnit Extensions.pax
+++ b/Core/Contributions/Camp Smalltalk/SUnit/OA SUnit Extensions.pax
@@ -10,6 +10,7 @@ Extensions to the basic SUnit unit testing packages, including the ability to te
 package methodNames
 	add: #TestCase -> #<=;
 	add: #TestCase -> #=;
+	add: #TestCase -> #assert:equals:description:;
 	add: #TestCase -> #assert:identicalTo:;
 	add: #TestCase -> #assert:isKindOf:;
 	add: #TestCase -> #assert:sameAs:;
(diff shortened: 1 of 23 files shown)


blairmcg:
I've noticed one problem, which is that if drags an item out of the list in the same gesture (i.e. mouse down on new selection, then drag out) that the selection does not update correctly.

blairmcg:
> I've noticed one problem, which is that if drags an item out of the list in the same gesture (i.e. mouse down on new selection, then drag out) that the selection does not update correctly.

Fixed in latest

PR #927, opened 2020-06-07
Running Database tests dirties the Northwind.mdb file

blairmcg:
Copy the test resource Northwind.mdb to temp dir before running the tests
against it to avoid creating uncommitted changes in the git workspace


blairmcg:
@fxgallego, I'll abandon this if you are proceeding with your refactoring.

Issue #780, opened 2019-05-20
Include Database Connections tests in the CI build


blairmcg:
The Database Connection tests are limited, and not even loaded with the regular tests, let alone executed. They don't have much value if they are not run, so we should add them in to the build, especially as @rko281 has done some work to improve them as part of fixing #768 and #778 recently.

This will mean including a Northwind.mdb in the repository, but I think that will be fine if added to Git LFS. We'll also need to ensure the necessary pre-requisites are available on the build machine, and degrade gracefully if not.

fxgallego:
I'm looking at this.
The tests run well with some minor adjustments.
I'd like to run the tests against Access and SQL Server.

I have a question about this test:
```
testViewsQuery
	| views |
	views := connection views.
	self 
		assert: (views noDifference: (self isAccessDB 
						ifTrue: 
							[#('Alphabetical List of Products' 'Category Sales for 1997' 'Current Product List' 'Customers and Suppliers by City' 'Employee Sales by Country' 'Invoices' 'Invoices Filter' 'Order Details Extended' 'Order Subtotals' 'Orders Qry' 'Product Sales for 1997' 'Products Above Average Price' 'Products by Category' 'Quarterly Orders' 'Quarterly Orders by Product' 'Sales by Category' 'Sales by Year' 'Ten Most Expensive Products')]
						ifFalse: 
							[#('Alphabetical list of products' 'Category Sales for 1997' 'Current Product List' 'Customer and Suppliers by City' 'Invoices' 'Order Details Extended' 'Order Subtotals' 'Orders Qry' 'Product Sales for 1997' 'Products Above Average Price' 'Products by Category' 'Quarterly Orders' 'Sales by Category' 'Sales Totals by Amount' 'Summary of Sales by Quarter' 'Summary of Sales by Year' 'sysconstraints' 'syssegments')]))
```
I've got `Northwind.mdb` but with other contents, for example: 'Category Sales for 1995', causing this test to fail. Do you have at hand the same database used in this tests? I can't find it. Anyway it's almost the same so I only have to repair this test.

blairmcg:
I've added as an LFS object in the original location referred to by the tests (which is in the package folder).
All the tests passed when I ran them against it.

fxgallego:
[...]

PR #930, opened 2020-06-09
Include Database Connections tests in the CI build #780

fxgallego:


fxgallego:
I think that it should build ok on master. The check failed because I made a mistake selecting the destination branch.

blairmcg:
Let's try close and reopen to see if it rebuilds

.gitattributes
```diff
@@ -16,3 +16,5 @@
 SciLexer.dll filter=lfs diff=lfs merge=lfs -text
 *.psd filter=lfs diff=lfs merge=lfs -text
 *.FH10 filter=lfs diff=lfs merge=lfs -text
+*.sql filter=lfs diff=lfs merge=lfs -text
```
blairmcg, line 19: Only the .mdb entry should be in this file. 
I forgot to add this .gitattributes file to the commit when I put the mdb into the repo as an LFS file, so that may be why you've ended up committing it.

fxgallego:
OK Removed mdb file. I was avoiding it for some time but it found the way to go in the commit.
Also removed lfs filter

PR #951, opened 2020-06-17
Add package class/inst properties to govern class pre-declaration on load

blairmcg:
Forward declaration of classes has a significant effect on package load times, but is historically what Dolphin has done. I recently disabled this unless packages contained source globals (e.g. pool constants dictionaries) but the effect is to break loading of code ported from Pharo that uses the idiom of representing pool dictionaries as classes. Classes as pools may be referenced before they are defined, and the hierarchy ordering of classes does not help. In order to re-enable scenarios that worked
previously, but which now do not, but still provide options to disable pre-declaration when it is known not to be required, this commit adds properties on the class and instance sides of Package. The class side setting controls the default choice when there is no per-package setting.
The default for this default is true, to behave as before (this may be changed in future). The setting can also be applied to individual packages and once set is persisted in the .pax file and respected regardless of the overall setting.
A 2nd commit sets the property to false (do not pre-declare) on all of the system packages, and some of the contributed packages. This isn't an indication that the other contributed packages need pre-declaration, they are just packages I try to avoid changing on the whole, perhaps because they are actively owned by the contributor.

### DIFF
--- a/Core/Contributions/Camp Smalltalk/ANSI/Tests/CS ANSI Tests.pax
+++ b/Core/Contributions/Camp Smalltalk/ANSI/Tests/CS ANSI Tests.pax
@@ -1,6 +1,7 @@
 ﻿| package |
 package := Package name: 'CS ANSI Tests'.
 package paxVersion: 1;
+	preDeclareClassesOnLoad: false;
 	basicComment: ''.
 
 package basicScriptAt: #postinstall put: '| locator |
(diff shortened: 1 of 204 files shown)


blairmcg:
@rko281, you should find that this change means you can remove your workaround for classes used as pool dictionaries now.
@fxgallego, as I've removed the optimisation of skipping pre-declaration when there are no source globals you will probably want to put a `Package preDeclareClassesOnLoad: false` at the top of your load script. You can also apply a permanent change to your packages by enumerating through all your packages and setting the same property to false. This will mark the packages as dirty, and when filed out they will record the setting. Obviously that will create a lot of source code churn, but you have the option.
In the next major release I will probably change the default to true (predeclare) to false (don't predeclare).

fxgallego:
@blairmcg ok, thanks no problem.

Issue #955, opened 2020-06-19
Unable to save edits to getContentsBlock in Enhanced list view


jgfoster:
**Describe the bug**
Unable to save code edits in View Composer aspect editor.

**To Reproduce**
1. Create a subclass of Shell
2. Create a Default View
3. Add a ListPresenter.Enhanced list view
4. Expand columnsList, first column
5. Double-click getContentsBlock
6. Enter a block expression: [:each | each first]
7. Save with green check button
8. Content stays as nil

**Expected behavior**
Code saved.

**Screenshots**
![Screen Shot 2020-06-18 at 10 37 35 PM](https://user-images.githubusercontent.com/1577872/85100179-56fef300-b1b4-11ea-9fd3-7e1e24221b2a.png)
**Please complete the following information):**
 - OS: Windows 10

**Additional context**
I've verified the problem back to Sept 11, 2019 (the first build to support VS2019). 

It has been a while since I've edited an Enhanced list view, but I think I'm doing the right things.

jgfoster:
I wonder if this is related to the fact that ValueWorkspace "is a somewhat unusual MVP triad; the ValueWorkspace model is not connected directly to the view. Instead, the view is redirected (via #viewModel) to another model that holds the textual rendition of the actual model." 

In `AspectEditorDialog>>apply`, the following line changes things so that the text entered becomes a String instead of leaving things as `nil`:

```
    "subjectModel value: aspectPresenterInDialog model value."
     subjectModel value: aspectPresenterInDialog viewModel value.
```

This is an improvement, but it eventually needs to be a block of code, not just a string.

blairmcg:
This bug has been around (and annoying me) for a while, I just haven't got around to logging it. Part of the reason for my laziness is that there is a workaround. If you first "Accept" in the workspace (using the menu) before pressing the OK tick, it should work. The workaround does at least allow the edit to be completed, although most of the time I end up doing it more than once after losing the change the first time because I forget that this doesn't work properly.

I'm not sure whether this ever worked correctly with the revised in-place editing in the VC that was put in for 7.0, or whether it was something that got broken later. There was a related problem that I think perhaps you might have logged James, and which I fixed but which may have caused this problem.

blairmcg:
Actually it was #575, and really very similar, so it must have regressed.

blairmcg:
Fixed by #956

Issue #575, opened 2018-07-05
Problems with saving code block in view composer


jgfoster:
For example, with a ListView->columnList->item->sortBlock, I can edit the code but if I click the green checkbox the edit is not saved. If I type Ctrl+S, it inserts a funny character. The only way to save is to select Accept from the context menu, then close the text field as a separate step.

PR #956, opened 2020-06-19
Aspect class>>block: seems happier with a defaultPresenterBlock

jgfoster:
Possible fix for https://github.com/dolphinsmalltalk/Dolphin/issues/955. I have no idea if this is right, but at least is saves the edits!

Thanks for the reminder of the work-around and the earlier report. There was something familiar about this.

### DIFF
--- a/Core/Object Arts/Dolphin/IDE/Base/Aspect.cls
+++ b/Core/Object Arts/Dolphin/IDE/Base/Aspect.cls
@@ -349,7 +349,7 @@ block: anAspectSymbol
 	for some kind of <valuable> (not necessarily a block).
 	The aspect is editable using a <ValueWorkspace> presenter."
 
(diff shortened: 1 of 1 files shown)


blairmcg:
I think it probably is the right fix. In 8c39b9b613328bd2dc535de95591c08ed047a0c6 for #575 I changed the code from passing a default block to passing nil in order to detect cases where the value workspace should be used and to set a flag because of its special behaviour. Then in a623d5f56e1587aeb1a3b5d7e6a3685ca473dfef for #596, I changed it again so that nil meant allow the use of the value workspace for displaying the object, but not for editing. Most other places that passed nil were changed to pass the defaultPresenterBlock, but it looks like this case was overlooked. This stuff really needs some tests.

PR #961, opened 2020-06-30
Make setters on OPENFILENAMEW nil-safe

daniels220:
Used to be that they just assigned the value, but now that they send #asUtf16String first they need to check for nil.

Or, would it be better to add `UndefinedObject>>asUtf16String`, with a comment explaining that this is an incredibly common pattern in `*W` struct classes? Or, make a separate `#asUtf16StringOrNil` or the like?

### DIFF
--- a/Core/Object Arts/Dolphin/MVP/Dialogs/Common/OPENFILENAMEW.cls
+++ b/Core/Object Arts/Dolphin/MVP/Dialogs/Common/OPENFILENAMEW.cls
@@ -52,8 +52,8 @@ defaultExtension: aString
 	"Set the default extension for the file open/save.
 	We store down the String in an instance variable to prevent it being GC'd."
 
-	defExt := aString asUtf16String.
-	self lpstrDefExt: defExt!
+	defExt := aString ifNotNil: [:a | a asUtf16String].
+	self lpstrDefExt: defExt.!
 
 dwSize: anInteger
 	"Set the receiver's 'dwSize' field to the value of the argument, anInteger"
@@ -69,15 +69,15 @@ fileName: aString
 	"Set the file name to be opened/saved.
 	We store down the String in an instance variable to prevent it being GC'd."
 
-	fileName := aString asUtf16String.
-	self lpstrFile: fileName!
+	fileName := aString ifNotNil: [:a | a asUtf16String].
+	self lpstrFile: fileName.!
 
 fileTypes: aString
 	"Set the file type filter to be used for the the common file dialog.
 	We store down the String in an instance variable to prevent it being GC'd."
 
-	filter := aString asUtf16String.
-	self lpstrFilter: filter!
+	filter := aString ifNotNil: [:a | a asUtf16String].
+	self lpstrFilter: filter.!
 
 flags
 	"Answer the <Integer> value of the receiver's 'flags' field."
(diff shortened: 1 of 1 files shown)


blairmcg:
What is the scenario that requires a need to check for nil? `OPENFILENAMEW` is really an internal detail of the `FileDialog` to support calling the underlying API. I can't seen any obvious reason to pass it nils instead of strings. If a parameter is not being specified, then there is no reason to set it. If any change is needed to support usage scenarios then it is more likely required in `FileDialog` and subclasses, but you will have to explain why these parameters would ever be nil when FileDialog initializes them to strings and its corresponding setter methods expect strings as their arguments.

daniels220:
I ran across it setting up a FileOpenDialog from a domain model, which may or may not have an existing filename associated with it. So, when I pass this existing filename to `FileOpenDialog>>value:` (and from there to `OPENFILENAMEW>>fileName:`), it may be a nil, and my code really doesn't care about that since it's about to get a new filename anyway. Prior to UTF support this worked fine—the underlying Win32 API is happy to accept a NULL in that slot, it seems.

blairmcg:
`FileDialog` already handles a nil value, e.g. try `(FileOpenDialog on: nil asValue) showModal`, so I can't see why that would be a problem.

A nil value is allowed for in the `prepareStruct` method, from which you can also see that the filename buffer is never null. It is up to the `FileDialog` classes to initialize the structure correctly. The buffer in the OPENFILENAMEW must not be null when opening or saving a file, as otherwise there is nowhere for Windows to pass back the chosen file name. The buffer can be initially empty, but never null. Therefore it doesn't makes sense for `OPENFILENAMEW>>fileName:` to accept a nil parameter. 


```smalltalk
FileDialog>>prepareStruct
prepareStruct
	"Private - Initialize the parameter structure to be passed to the Win32 API associated with this common dialog, immediately prior to the call to that API."

	| buf defext |
	buf := Utf16String newFixed: self maxFileCount * self averageFilenameLength + File maxPath.
	self value
		ifNotNil: 
			[:filename |
			| utf16 |
			utf16 := filename asUtf16String.
			buf
				replaceFrom: 1
				to: utf16 size
				with: utf16].
	(self filterIndex = 0 and: [(defext := self defaultExtension) notNil])
		ifTrue: [self filterIndex: (self fileTypes findFirst: [:pair | pair last contains: defext])].
	self winStruct
		fileTypes: (self fileTypesStringFromSpecs: self fileTypes);
		nMaxFile: buf size;
		fileName: buf;
		maskIn: self style
```

daniels220:
Ah, my mistake—the issue I had was with with `defaultExtension`, which is just passed straight through—and that definitely *is* valid to be NULL. Looking more closely, seems like the nil-checks in `fileName:` and `fileTypes:` aren't appropriate as those are handled internally by `FileDialog`, but `defaultExtension:`, `initialDirectory:` and `title:` are just passed through and accept NULL.

blairmcg:
I suppose it doesn't hurt, but I find it a bit untidy to zero out the values again since "nothing" is already the value. A better solution in the original design would have been to store all the values in instance variables and then set everything up in the `prepareStruct` method, rather than passing stuff directly through like it does. At some point I might rework the whole hierarchy so that the structure is only created as a temporary for the duration of the modal operation. Only the FindDialogs can operate in a modeless fashion.

daniels220:
That makes sense. I'm fine with closing this PR and just patching it locally for my needs.

PR #963, opened 2020-06-30
Fix String>>copyWithout: to correctly recognize non-ASCII characters in a Utf8String (fixes #957)

### LINKED ISSUE #957: Inconsistent handling of Strings as characters vs bytes
`'abc°def' copyWithout: $°` yields an unmodified string—the degree symbol is not removed. If I convert the string to ANSI first it works as expected. Looking at the implementation of `copyWithout:` and tracing from there, it seems clear that `String>>do:` is a byte-wise iteration, and ends up comparing the ° to each of the two bytes that make it up in UTF-8, which it doesn't match, and so nothing is removed.

ReadStream, by contrast, handles strings character-wise, so if I replace `self do:` with `self readStream do:` in `SequenceableCollection>>copyWithout:` (or add an override that behaves this way on `String`), the problem goes away. This seems like a more general issue of inconsistency/lack of clarity about when a String is treated as an array of bytes-that-are-accessed-as-characters, and when it is treated as an array of Characters-that-may-not-all-be-the-same-size.

daniels220:
Override on UtfEncodedString to iterate using a ReadStream rather than directly.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/UtfEncodedString.cls
+++ b/Core/Object Arts/Dolphin/Base/UtfEncodedString.cls
@@ -61,6 +61,18 @@ copyWith: newElement
 
 	^self, (self class with: newElement)!
 
+copyWithout: aCharacter
+	"Answer a <String> which is a copy of the receiver, but in
+	which all occurrences of the <Character> aCharacter have been removed.
+
+	N.B. Use a ReadStream to process whole UTF-8 codepoints at once--otherwise
+	we will be unable to recognize non-ASCII characters in a UTF-8 string."
+
+	| aStream |
+	aStream := (self copyLikeOfSize: self copySize) writeStream.
+	self readStream do: [:element | element = aCharacter ifFalse: [aStream nextPut: element]].
+	^aStream contents.!
+
 decodeFrom: aReadStream upTo: anObject
 	"Private - Answer the future sequence values of the <ReadStream>, aReadStream, up to but not including, the <Object>, anObject.
 	The stream is left positioned after anObject. If there are no occurrences of anObject in the future sequence values of the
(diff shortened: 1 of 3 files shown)


Core/Object Arts/Dolphin/Base/UtfEncodedString.cls
```diff
@@ -61,6 +61,18 @@ copyWith: newElement
...
 
+copyWithout: aCharacter
+	"Answer a <String> which is a copy of the receiver, but in
+	which all occurrences of the <Character> aCharacter have been removed.
+
+	N.B. Use a ReadStream to process whole UTF-8 codepoints at once--otherwise
+	we will be unable to recognize non-ASCII characters in a UTF-8 string."
+
+	| aStream |
+	aStream := (self copyLikeOfSize: self copySize) writeStream.
```
blairmcg, line 72: This doesn't need to be so generic at this point in the hierarchy.  It could just be the simpler `self class writeStream: self size` here. But it doesn't matter, I'll change that at some later point.

Core/Object Arts/Dolphin/Base/UtfEncodedString.cls
```diff
@@ -142,6 +154,7 @@ skipEncodingMarkerFrom: aPositionableStream
 !UtfEncodedString categoriesFor: #asArray!converting!public! !
 !UtfEncodedString categoriesFor: #bom!constants!encode/decode!private! !
 !UtfEncodedString categoriesFor: #copyWith:!copying!public! !
+!UtfEncodedString categoriesFor: #copyWithout:!converting!public! !
```
blairmcg, line 157: Should be `copying` rather than `converting`

PR #964, opened 2020-06-30
SYSTEMTIME>>asTimeStamp should clearly answer a TimeStamp, not a Time

daniels220:
Just a little something I stumbled across...


Core/Object Arts/Dolphin/Base/Deprecated/Dolphin Base (Deprecated).pax
```diff
@@ -1102,7 +1102,7 @@ unicodeClass
 
 asTimeStamp
 	Notification deprecated.
-	^self asDateAndTime asTime! !
+	^self asDateAndTime asTimeStamp.! !
```
blairmcg, line 1105: I'm not sure this is worth fixing, as it is deprecated. I'm even less sure it is worth fixing without a test. Still...

PR #977, opened 2020-07-10
A literal String should not count as "referring to" an equal Symbol.

daniels220:
Different classes of String should still be allowed to "refer to" each other, so the Object implementation isn't appropriate.

I also noticed a similar situation in the Refactoring Browser code, in `RBMethod>>literal:containsReferenceTo:`. Since there the argument clearly must be a `Symbol`, I just changed it to an `==` check. I'm not sure where to test it, though...by the time it works its way through to an actual scenario, I think it would be hideously complicated to set up a test. Any suggestions?

### DIFF
--- a/Core/Object Arts/Dolphin/Base/String.cls
+++ b/Core/Object Arts/Dolphin/Base/String.cls
@@ -878,11 +878,11 @@ printOn: aStream
 				ifFalse: [(aStream nextPut: ch) == $' ifTrue: [aStream nextPut: $']]].
 	aStream nextPut: $'!
 
-refersToLiteral: anObject 
+refersToLiteral: anObject
 	"Private - Answer whether the receiver is a reference to the literal argument.
 	This assumes that the receiver is in the role of a literal."
 
-	^self = anObject!
+	^anObject isSymbol not and: [self = anObject].!
 
 replaceBytesOf: aByteObject from: start to: stop startingAt: fromStart
 	"Private - Standard method for transfering bytes from one variable
--- a/Core/Object Arts/Dolphin/IDE/Base/CompilerTest.cls
+++ b/Core/Object Arts/Dolphin/IDE/Base/CompilerTest.cls
@@ -1086,6 +1086,18 @@ testSpecialLiteralsAsUnarySelectors
 				raise: self compilationErrorClass.
 			self denyIsNil: method]!
 
+testStringDoesNotReferToSymbol
+	"Test that string literals in a method do not count as referring to a Symbol with the same contents,
+	now that they are #= to said Symbol.
+
+	N.B. they *should* still count as referring to other encodings of the same string."
+
+	| method |
+	method := (self compileExpression: '''abc'' size') method.
+	self assert: (method refersToLiteral: 'abc' asAnsiString).
+	self assert: (method refersToLiteral: 'abc' asUtf8String).
+	self deny: (method refersToLiteral: #abc)!
+
 testSymbolScanning
 	#('#a' '#a1' '#a:' '#a1:' '#a:b:' '#a_:_:' '#a1:b2:' '#|' '#||' '#|||' '#_' '#_a' '#-' '#--' '#''£''' '#''你好''' '#''🐬''') do: 
 			[:each |
@@ -1546,6 +1558,7 @@ verifyTextMapsOf: each
 !CompilerTest categoriesFor: #testScope!public!unit tests! !
 !CompilerTest categoriesFor: #testSharedOuterTemps!public!unit tests! !
 !CompilerTest categoriesFor: #testSpecialLiteralsAsUnarySelectors!public! !
+!CompilerTest categoriesFor: #testStringDoesNotReferToSymbol!public!unit tests! !
 !CompilerTest categoriesFor: #testSymbolScanning!public!unit tests! !
 !CompilerTest categoriesFor: #testTempMaps!public!unit tests! !
 !CompilerTest categoriesFor: #testTempPairing!public!unit tests! !
--- a/Core/Object Arts/Dolphin/IDE/Base/CompilerTest.cls
+++ b/Core/Object Arts/Dolphin/IDE/Base/CompilerTest.cls
@@ -1086,18 +1086,6 @@ testSpecialLiteralsAsUnarySelectors
 				raise: self compilationErrorClass.
 			self denyIsNil: method]!
 
-testStringDoesNotReferToSymbol
-	"Test that string literals in a method do not count as referring to a Symbol with the same contents,
-	now that they are #= to said Symbol.
-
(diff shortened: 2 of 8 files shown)


blairmcg:
> I also noticed a similar situation in the Refactoring Browser code, in `RBMethod>>literal:containsReferenceTo:`. Since there the argument clearly must be a `Symbol`, I just changed it to an `==` check. I'm not sure where to test it, though...by the time it works its way through to an actual scenario, I think it would be hideously complicated to set up a test. Any suggestions?

Just test it as directly as you can. There is no RBMethodTest, so I'd add one.


```smalltalk
RBMethodTest>>testLiteralStringsDoNotMatchSymbols

    | class method |
    class := RBClass existingNamed: #RefactoryTestDataApp.
    method := class methodFor: #returnInEnsure.
    self deny: (method refersToSymbol: #asdf).
```

I'm not sure your change will be sufficient to make this test work though.

Core/Object Arts/Dolphin/Base/String.cls
```diff
@@ -878,11 +878,11 @@ printOn: aStream
 				ifFalse: [(aStream nextPut: ch) == $' ifTrue: [aStream nextPut: $']]].
 	aStream nextPut: $'!
 
-refersToLiteral: anObject 
+refersToLiteral: anObject
 	"Private - Answer whether the receiver is a reference to the literal argument.
 	This assumes that the receiver is in the role of a literal."
 
-	^self = anObject!
+	^anObject isSymbol not and: [self = anObject].!
```
blairmcg, line 885: I'm not keen on the type test, but I suppose it can't be avoided without introducing a double-dispatch, which would be overkill for this. Oh well...

Core/Object Arts/Dolphin/IDE/Base/CompilerTest.cls
```diff
@@ -1086,6 +1086,18 @@ testSpecialLiteralsAsUnarySelectors
 				raise: self compilationErrorClass.
 			self denyIsNil: method]!
 
+testStringDoesNotReferToSymbol
```
blairmcg, line 1089: A test for #refersToLiteral: between symbols and strings does not belong here as it is nothing to do with the compiler. You don't need to involve a method at all - the logic of searching through the literals is not relevant to what you are trying to test. I would just keep it simple and put it in StringTest/SymbolTest invoking String>>refersToLiteral: and Symbol>>refersToLiteral: directly in the necessary combinations.

daniels220:
Re: RBMethod, you're right that my change isn't enough, for an interesting reason—a literal node containing a Symbol is being considered equal to one containing a String, so `#literal:containsReferenceTo:` is never being hit. In turn this is because `Utf8String` and `Symbol` both answer `Utf8String` as their `species`, and:

```Smalltalk
StLiteralValueNode>>= anObject 
	^super = anObject and: 
			[self value = anObject value 
				and: [self value species == anObject value species]]
```

Long ago the `=` check would have failed, but now that it passes... Off the top of my head, it seems like a `Symbol` and a `Utf8String` are more different than a `Utf8String` and an `AnsiString`, so it's weird that they are the "same species" while the Utf8 and Ansi strings are different species. But I see that this is very much intentional—I imagine because `#species` is used by e.g. `#select:` to create a new collection, and that must be a String of some kind, not a Symbol.

So...any thoughts on the best way to ensure that (a) literal value nodes containing the same string in different encodings are considered equal, while (b) one containing a string is not equal to one containing a same-contents symbol?

Core/Object Arts/Dolphin/IDE/Base/CompilerTest.cls
daniels220: Yeah, good point, done.

blairmcg:
> Long ago the `=` check would have failed, but now that it passes... Off the top of my head, it seems like a `Symbol` and a `Utf8String` are more different than a `Utf8String` and an `AnsiString`, so it's weird that they are the "same species" while the Utf8 and Ansi strings are different species. But I see that this is very much intentional—I imagine because `#species` is used by e.g. `#select:` to create a new collection, and that must be a String of some kind, not a Symbol.
> 
> So...any thoughts on the best way to ensure that (a) literal value nodes containing the same string in different encodings are considered equal, while (b) one containing a string is not equal to one containing a same-contents symbol?

Hmmm, that is tricky. The problem is that `species` has overloaded meanings. It really ought to mean species in the real sense of the word, but there are other uses such as the one you mention of using it to implement `select:` in a generic way (which I think comes from the original Smalltalk class library). Frankly that is rather stretching the meaning and then quandaries such as this arise. I can't immediately see a way to resolve it without addressing the misuse in the collection hierarchy, which is not something I'd want to do in a minor increment.

How much does it matter? If it's an immediate problem we can log an issue to address later. The test can be left in, but marked as skipped.

daniels220:
Agreed about the overloaded meaning of `species`. I don't have an immediate need to have this fixed, but it does seem like a potentially-serious problem that `(RBLiteralValueNode value: 'foo') = (RBLiteralValueNode: #foo)`, and thus that a Code Rewriter search containing a literal symbol will also match a method with an equivalent literal string. So, I made another issue for that (#978).

I actually can make this particular test pass *without* fixing the underlying issue, though—AFAICT, the direct search for the bare symbol literal is redundant, as the more general search of any literal with `literal:containsReferenceTo:` will match the bare-symbol case. If that seems correct to you, it seems worth doing...

Core/Object Arts/Dolphin/Base/String.cls
daniels220: Right, yeah, it bothered me too, but I couldn't see a better way.

brunobuzzi:
(RBLiteralValueNode value: 'foo') = (RBLiteralValueNode value: #foo)
Dolphin: false
GemStone/S: false
VA Smalltalk: false
Visual Works 8.3: false
Pharo 8.0: true

blairmcg:
The fact that the underlying values of two literal nodes are equal does not necessarily mean that the literal nodes are equal:

```Smalltalk
(RBLiteralValueNode value: 1) = (RBLiteralValueNode value: 1.0) "false - correct, these nodes should not be considered equal"
1 = 1.0 "true - but of course the values themselves are equal"
```

> (RBLiteralValueNode value: 'foo') = (RBLiteralValueNode value: #foo)
>...
> Pharo 8.0: true

Which is incorrect. It should always be false. Dolphin also gets it wrong and for similar reasons (hence the discussion above). Your example gives the correct result in Dolphin 7.1 because it compiles plain ASCII strings as a different string encoding than is used for Symbols. If you'd instead written:

```Smalltalk
(RBLiteralValueNode value: '🐬') = (RBLiteralValueNode value: #'🐬')
```

Then Dolphin would answer `true`, which is incorrect, even though `'🐬' = #'🐬'` is true.

PR #983, opened 2020-07-17
Fix #957

### LINKED ISSUE #957: Inconsistent handling of Strings as characters vs bytes

blairmcg:
This is a more radical adaption of UTFEncodedString into the collection hierarchy so that most methods work exclusively with code points rather than code units. The rule is that indices used to access into a String are always code unit indices. Sizes (with the exception of the String size itself, which is a code unit count) are expressed as the number of code points. Although this does seem a bit odd on the face of it, it generally works as long as one doesn't assume that the interval between two indices is the same as the number of characters in that interval.
This does change the behaviour of some methods that may have previously returned surrogate characters, e.g. `first` and `last`, and also means that `do:` and other enumerations are over complete Characters.


blairmcg:
@fxgallego, I changed your `urlEncoded` method.

Issue #990, opened 2020-08-11
Stripper fail if a referenced resource has been deleted


brunobuzzi:
**Describe the bug**
The Stripper fail if a referenced resource has been deleted.
The problem can be tracked to `ResourceIdentifier>>hiddenObjects`. When `self resource` answer `nil` the Stripper halt abruptly with not further information which make hard to track the problem.
**To Reproduce**
1. Create PresenterA (with a View)
2. Create ShellA (referencing PresenterA view)
3. Delete resource (view) method of PresenterA (Refactoring Remove will warn you about the reference but you can do a mistake and use Delete instead)
4. Try to Deploy  (it will fail)

**Expected behavior**
An clear Exception will be very useful

brunobuzzi:
I have fixed this i will send the PR in a few hours.

brunobuzzi:
Fixed with:
https://github.com/dolphinsmalltalk/Dolphin/pull/991

PR #991, opened 2020-08-11
fix issue #990

brunobuzzi:
This modification prompt a message with the offending resource so it will be easy to track the problem.
Without this message it can take a lot of time to find out the offending resource.

### DIFF
--- a/Core/Object Arts/Dolphin/MVP/ResourceIdentifierTest.cls
+++ b/Core/Object Arts/Dolphin/MVP/ResourceIdentifierTest.cls
@@ -22,6 +22,18 @@ createResourceNamed: aString for: aView
 	^rid
 !
 
+createResourceNamed: aString for: aView on: aPresenterClass
+	| filer rid resource |
+
+	filer := STLOutFiler on: Array writeStream.
+	filer nextPut: aView.
+	aView destroy.
+	resource := filer stream contents.
+	rid := ResourceIdentifier class: aPresenterClass name: aString.
+	rid assign: resource.
+	^rid
+!
+
 setUp
 	| shellView filer |
 	super setUp.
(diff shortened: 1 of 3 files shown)


brunobuzzi:
> 
> 
> Bruno, per the contribution guidance in the readme it requires a test.
@blairmcg ok. I will create some test and add it to the PR.

brunobuzzi:
@blairmcg i added a test.
Without this fix when producing a EXE the following popup is displayed:
issue-990
Is difficult to find what happen...

With the FIX the following popup is displayed:
issue-990-fix
It show the offending resource and it point out the problem.

This error can happen when removing resources without the Refactoring menu (using delete directly).
For now it answer `self error: ....` but maybe `MissingResourceException` can be used ...

brunobuzzi:
It seems that 2 tests failed but has nothing to do with this PR:

`ResourceIdentifierTest>>testReferenceViewRemoved ... PASSED in 93.8 ms`

ERRORS:
```
ArrayPointerFieldTest>>testGenerateAccessorsForSizedString
ArrayPointerFieldTest>>testGenerateAccessorsForSizeIsExpression
```

Core/Object Arts/Dolphin/MVP/ResourceIdentifierTest.cls
```diff
@@ -10,6 +10,18 @@ ResourceIdentifierTest comment: ''!
...
 !ResourceIdentifierTest methodsFor!
 
+createResourceNamed: aString for: aView
+	| filer rid resource |
+
+	filer := STLOutFiler on: Array writeStream.
+	filer nextPut: aView.
+	aView destroy.
+	resource := filer stream contents.
+	rid := ResourceIdentifier class: Presenter name: aString.
```
blairmcg, line 20: Unfortunately this is going to leave the Presenter class and base package marked as dirty when people run the test. You should be able to install the resource into an anonymous class created by the test method to avoid this problem.

brunobuzzi: @blairmcg  
Now the Test use an anonymous subclass of Presenter.

PR #1000, opened 2020-08-24
Method History Browser improvements

### LINKED ISSUE #999: Improve Method History Browser performance
The Method History Browser is very slow in my current working image (around 60MB changes file) - 10 seconds to open, then another 10 seconds as it refreshes when I reinstall an earlier method version (this wasn't the case in D7.0). The core of the slowdown appears to be the send of `nextAvailable:` in `ChunkFileMethodScanner>>positionsOf:in:firstOnly:` as it reads and decodes the requested number of UTF8 characters.

I initially implemented a modification whereby ChunkFileMethodScanner instead reads a specified number of bytes via `basicNextAvailable:` plus any trailing UTF8 character units (the exact number of characters read is unimportant; it's just part of the local paging scheme). The result is a String of valid character points <= the requested number, and a much improved performance.

However a much simpler solution with similar performance is to remove the local paging scheme and simply read the entire file in one hit. I'm wondering whether this would be acceptable - the likely maximum size of a changes file (100MB?) isn't _that_ large for current machines.

rko281:
Speed up change file scanning; reuse existing source when restoring methods. Plus associated functional and cosmetic changes. 

Closes #999.


blairmcg (changes requested):
John, thanks for the contribution, which is a good improvement overall. I tested it out and it works well except that when a method history is open it no longer collapses identical adjacent revisions made after it is opened. If closed and re-opened it does still do so though.
Per the contribution guidance in the read me, it also requires some tests. I would say at least of the scanning change that is the original motivation, in particular checking chunks starting on or just straddling (at either end) the boundary between blocks. It will probably be necessary to make the block size parameterisable in some way to test that easily. Also the bug with not coalescing identical adjacent versions suggests a need for a test there too.

rko281:
Sure, no problem. Will resubmit with a bugfix and tests.

PR #1002, opened 2020-08-27
Fix DiffBrowser "character" comparison mode.

daniels220:
Didn't actually work--the code ended up using the same items as "word" mode due to a control-flow mistake.


blairmcg (changes requested):
Per the contribution guidance in the readme,  tests covering any changes (or a very compelling explanation for the lack of) are required.

daniels220:
Right, fair. Sorry, many of these PRs are for things I did a long time ago, before Dolphin was open-source, and as a result before the tests were even available, and I forget to go back and add tests now that they are.

PR #1004, opened 2020-08-27
Replace #unadornedText with proper use of Win32 functions in OwnerDrawnButton

daniels220:
All drawing functions used accept the DSS_HIDEPREFIX flag, make use of it instead of stripping out `&`s ourselves.

### DIFF
--- a/Core/Contributions/Solutions Software/OwnerDrawnButton.cls
+++ b/Core/Contributions/Solutions Software/OwnerDrawnButton.cls
@@ -76,12 +76,6 @@ displayState
 displayState: anObject
 	displayState := anObject!
 
-displayText
-
-	^self shouldShowKeyboardAccelerators
-		ifTrue: [self text]
-		ifFalse: [self unadornedText]!
-
 drawOn: aCanvas
 
 	self subclassResponsibility!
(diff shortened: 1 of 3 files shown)


blairmcg (changes requested):
Seems like a reasonable change to me, but as this is part of a live contribution from @rko281, I will defer to John as to whether he wants to accept it. Even if he is happy, it still requires test coverage.

rko281:
Fine with me - don't think I was aware of the flag hence the manual implementation.

daniels220:
Any suggestion on how to test it? If this code were to break, the only thing that would be noticeable is incorrect rendering on the screen—nothing with a return value or the like that I can check.

blairmcg:
Without looking in depth, I would generally agree that where the effects are purely visual testing can be hard and may not be justifiable. Where that is the case the PR should explain why as I don't necessarily have the time to work it out for myself. 
As this is an improvement rather than a bug fix as such, my main concern would be automatically detecting that the enhancement is not lost or broken in future. It would be reasonable to test just that the text does actually draw in the two modes, which I would think could be done using the same trick as I used in my recent PR #993. See the method `TabViewXPTest>>#testLabelWithAmpersand`.

PR #1006, opened 2020-08-27
Make sure newly-changed packages move to the top of the list if sorted by status

daniels220:
Straightforward enough to just re-sort the whole list, though it seems like there should be a way to tell the ListPresenter this directly rather than telling it "hey, be sorted the same way you already are". It also brings to mind that there doesn't seem to be an API for asking "which column, if any, is this ListPresenter currently sorted on?". That state is clearly held by the view somehow, in order to display the little arrow in the header...

### DIFF
--- a/Core/Object Arts/Dolphin/IDE/Base/PackageSelector.cls
+++ b/Core/Object Arts/Dolphin/IDE/Base/PackageSelector.cls
@@ -488,6 +488,7 @@ onPackageChanged: aPackage
 	"Private - The development system has marked aPackage as being changed,
 	update the package list to show any change in the appearance of the package."
 
+	self packagesPresenter beSorted: self packagesPresenter sortBlock.
 	self packagesPresenter view updateAll!
 
 onPackageChosen


Core/Object Arts/Dolphin/IDE/Base/PackageSelector.cls
```diff
@@ -488,6 +488,7 @@ onPackageChanged: aPackage
 	"Private - The development system has marked aPackage as being changed,
 	update the package list to show any change in the appearance of the package."
 
+	self packagesPresenter beSorted: self packagesPresenter sortBlock.
```
blairmcg, line 491: I'm not utterly convinced of the value of this - the sort will be unstable so the list will re-order every time a new package is marked as changed. It is also fairly heavyweight and there will be a double re-draw when it does resort (the subsequent `updateAll` is unnecessary in that case). There is good precedent (e.g. Windows Explorer) not to re-sort every time a new item is added, but rather expect the user to re-sort again should they wish to.
As I don't generally sort the package lists anyway and the change event only fires when a package is first marked as changed, this won't really have any negative effect for me, so I don't care enough to refuse it. However it will require a test, if only to detect it getting lost on a future mismerge.

blairmcg (changes requested):
T

PR #1007, opened 2020-08-27
Produce more consistent sortStrings for Dictionary and Set

daniels220:
Attempt to print the keys/values in sorted order, falling back to hash order if sorting fails


blairmcg (commented):
I don't agree with this change. The `storeOn:` mechanism is a simple persistence format with long established behaviour. Although it is source text, it isn't source code nor intended to be particularly readable. Imposing the cost of a sort in a low-level mechanism like this is not appropriate.

HenrikHoyer:
I agree with @blairmcg

PR #1027, opened 2020-10-07
Fix primitiveFailed error when closing the View Composer

daniels220:
The View Composer tries to redraw the adornment after every command, and if the command was e.g. `#exit` and the composer is now closed, this causes an error. I added a check of `#isOpen` before the redraw to prevent this.

For the test, I did what I've always done in my own work to simulate a button press or menu action—let me know if there's a cleaner or more accurate way. It does need to go through `#performCommand:`, as that's where the bug was.


Core/Object Arts/Dolphin/IDE/Base/ViewComposer.cls
```diff
@@ -78,15 +78,9 @@ adornmentRectanglesFor: aView
 	"Private - Answer a collection of rectangles that mark the edges of the secondary
```
blairmcg, line 78: This method is not used and could just be deleted. No need to worry about that now though.

PR #1046, opened 2020-10-14
Fix intermittent "common control call failed" error working with DateTimePicker (fixes #1045)

### LINKED ISSUE #1045: Intermittent "Common control call failed" error creating (or updating?) DateTimePicker
I'm seeing a very occasional "Common control call failed" error in `dtmSetSystemTime:lpSysTime:`, when a DateTimePicker's model is freshly assigned, and possibly also when its value changes. The error doesn't give any details, and I don't know enough about Win32, never mind the Common Controls, to have any idea if there are details available somewhere. The most common stacktrace looks something like:

```
DateTimePicker(ControlView)>>errorInCommonControlCall
DateTimePicker>>dtmSetSystemTime:lpSysTime:
...
DateTimePicker(View)>>onModelChanged
DateTimePicker(View)>>model:
DatePresenter(Presenter)>>connectView
DatePresenter(Presenter)>>view:
...
MyShell(Presenter)>>attachSubPresenterViews:
...
MyShell(Presenter)>>createView:
```

Any ideas (a) what the problem is and how to actually fix it, or failing that (b) how to get more info about the error, or failing that (c) any possible workaround (i.e. what would I have to do to retry with as little disruption as possible—is literally just retrying likely to work? Destroy/recreate the DateTimePicker first? Call `pumpMessages` (I hope not, that would be Wrong even if it worked...)?

daniels220:
Truncate milliseconds rather than rounding to ensure the value is in range 0..999, rather than occasionally setting 1000 which is invalid.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/DateAndTime.cls
+++ b/Core/Object Arts/Dolphin/Base/DateAndTime.cls
@@ -86,7 +86,7 @@ asSYSTEMTIME
 				wHour: self hour24;
 				wMinute: self minute;
 				wSecond: secs truncated;
-				wMilliseconds: (secs fractionPart * 1000) rounded;
+				wMilliseconds: (secs fractionPart * 1000) truncated;
 				yourself]!
 
 asUnixTime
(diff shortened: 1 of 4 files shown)


blairmcg:
@daniels220, I'll merge this, and then fix the pre-existing test that is failing, since it is just assuming the old rounding behaviour, rather than the new truncating behaviour.

PR #1048, opened 2020-11-04
Fix race condition in Delay>>wait when #signal is called at approximately the same time as #wait

daniels220:
Given two processes at different priorities—call them `background` (at relatively low priority) and `foreground` (at relatively high priority)—and a `Delay` object `some_delay`, it is possible for the following sequence to occur:

1. `background` calls `some_delay wait`, enters `Delay>>wait` and executes `calcResumptionTime`.
2. At some point between the return from `calcResumptionTime` and acquiring `Delay.AccessProtect`, `foreground` becomes runnable, preempts and calls `some_delay signal`. This sets `some_delay.resumptionTime` ivar to `nil`, but the `resumingAt` temp in `#wait` is still set.
3. Once `foreground` waits on something else (an overlapped I/O call, in the case where I encountered this), `background` acquires `Delay.AccessProtect` and ultimately tries to either `schedule` or `snooze`, both of which will fail because `resumptionTime` is nil.

My thinking is that, for this to occur, the call to `signal` must happen after the call to `wait`, so the more-correct thing is for the delay to ultimately not wait, and since the `signal` call will have left an unconsumed signal behind, all we need to do is not try to schedule/snooze the delay and everything will work out.

I have a reproducible test-case for this, but at the moment it involves proprietary application code—I haven't managed to trace the exact interaction of the various Processes and reproduce it using only dummy code and other Delays. If doing so is important, I can keep trying, but the mechanism seems clear enough...

### DIFF
--- a/Core/Object Arts/Dolphin/Base/Delay.cls
+++ b/Core/Object Arts/Dolphin/Base/Delay.cls
@@ -45,7 +45,7 @@ Example usage:
 	^self resumptionTime <= anotherDelay resumptionTime!
 
 basicCancel
-	^self == Current
+	self == Current
 		ifTrue: 
 			[self class
 				cancelTimer;
--- a/Core/Object Arts/Dolphin/Base/Delay.cls
+++ b/Core/Object Arts/Dolphin/Base/Delay.cls
@@ -185,20 +185,24 @@ terminateProcess
 	self cancel!
 
 wait
-	"Delay the active Process until the receiver's resumption time. The processes priority is temporarily raised to prevent it holding up the timing process when its critical  section - if we did not do this and the current active process has a low priority, then it might be preempted in its critical section, which will prevent the TimingProcess from operating correctly. The resumption time is calculated outside the critical section so that it is based, as near as possible, on the millisecondClockValue at the time the #wait message is sent. If it were calculated inside, then it would incorrectly include the time taken to acquire the AccessProtect Semaphore."
+	"Delay the active Process until the receiver's resumption time. The processes priority is temporarily raised to prevent it holding up the timing process when its critical  section - if we did not do this and the current active process has a low priority, then it might be preempted in its critical section, which will prevent the TimingProcess from operating correctly.
+
+	The resumption time is calculated outside the critical section so that it is based, as near as possible, on the microsecondClockValue at the time the #wait message is sent. If it were calculated inside, then it would incorrectly include the time taken to acquire the AccessProtect Semaphore. However, we must check again upon entering the critical section that it has not been cleared by a #signal in the interim--in this case we have nothing to do, similar to if we were scheduled for an amount of time less than the current resolution."
 
 	| resumingAt |
 	resumingAt := self calcResumptionTime.
-	AccessProtect critical: [
-		"If the new Delay will be the first to wakeup, schedule it"
-		Current isNil 
-			ifTrue: [self schedule]
-			ifFalse: [
-				resumingAt < Current resumptionTime
-					ifTrue: [ 
-						Current snooze.
-						self schedule]
-					ifFalse: [self snooze]]] atPriority: TimingProcess priority-1.
+	AccessProtect critical: 
+			[resumptionTime isNil ifFalse: 
+					["If the new Delay will be the first to wakeup, schedule it"
+					Current isNil
+						ifTrue: [self schedule]
+						ifFalse: 
+							[resumingAt < Current resumptionTime
+								ifTrue: 
+									[Current snooze.
+									self schedule]
+								ifFalse: [self snooze]]]]
+		atPriority: TimingProcess priority - 1.
 	"Now make the active Process wait"
 	waitSemaphore wait!
 
--- a/Core/Object Arts/Dolphin/Base/Delay.cls
+++ b/Core/Object Arts/Dolphin/Base/Delay.cls
@@ -44,6 +44,14 @@ Example usage:
 
 	^self resumptionTime <= anotherDelay resumptionTime!
 
+basicCancel
+	^self == Current
+		ifTrue: 
+			[self class
+				cancelTimer;
+				scheduleNext]
+		ifFalse: [Pending remove: self ifAbsent: []]!
+
 calcResumptionTime
 	"Private - Calculate and answer the value of the millisecond clock after which any Process
 	waiting on the receiver will be rescheduled. We could lazily evaluate this from the normal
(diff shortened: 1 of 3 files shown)


blairmcg (commented):
Firstly, just sending a change without discussion around an issue, and without a clear repro, and without any tests, is unlikely to be accepted. If the change affects a critical part of the system, it is extremely unlikely to be accepted. I'd have thought this was clear before, but if not it I hope it is now.

I would say that a repro is definitely required for any race condition as a starting point to assessing the problem and validating any proposed fix. Blind fixes would only be worth accepting where the value outweighs the risk. A PR is not the place for discussing the value/risk trade-off.

Without a repro we can't hope to add a test for the issue as described, but ideally I would want to have that to ensure the issue is properly fixed, and (even more importantly) remains so. Even race conditions can usually be simulated when they are understood.

Then there is the potential to break other scenarios to be considered. The issue described relates to sending the signal message to a Delay before it is completely waiting. This is not core in itself, which may explain why the bug has been there so long, but the change proposed is to the mainline code path for waiting on a Delay, i.e. impacting the core scenario. Given that there are no existing tests of that core scenario, I'd be reluctant to change this at all without closing that test gap, even for such an apparently innocuous change that looks to be correct. The general rule is that unchanged code without tests is grandfathered in. Old code becomes new code as soon as it is changed, and all new code must be covered adequately by new or existing tests.

Finally, I'm not convinced that the proposed fix removes the race conditions completely anyway. The unchanged `signal` method does not perform all of its actions within the critical section. The private `wakeup` method looks like it expects to be performed within the critical section, just as it is when the timing process calls it. If not, then there is still an unlikely but possible opportunity for the `resumptionTime` instance variable to be nil'd out even after the wait method has acquired the critical section and tested for nil but before the value is required to schedule the timer.

blairmcg:
> Without a repro we can't hope to add a test for the issue as described, but ideally I would want to have that to ensure the issue is properly fixed, and (even more importantly) remains so. Even race conditions can usually be simulated when they are understood.

This would appear to be a repro for the case you describe if I understand it correctly. Be sure to run it in a scratch image as the damage is difficult to repair.

```smalltalk
d := Delay forDuration: 1 seconds.
Delay.AccessProtect wait. "Ensure that the Delay wait will be blocked"
[d wait] fork. "Delay>>wait will calc a resumption time, then block"
[d signal] forkAt: Processor userInterruptPriority. "Will block on AccessProtect in #cancel too, but higher priority"
Delay.AccessProtect signal. "Allow the bug to unfold"
```

daniels220:
Okay, in the future I can make an associated issue, and mark PRs as drafts if they’re incomplete in some way (such as lacking tests)—I haven’t used that feature before and didn’t think of it here, but can do so in the future.

You make a good point about needing test coverage of the normal scenarios for Delay>>wait when making changes that might impact them. I’ve done my best to do so, but I have to say I’m not too confident in my ability to either (a) cover all the various scenarios, or (b) ensure that the tests I write are actually capable of failing in a useful way. Let me know what you think.

Your reproduction is very clever! It never occurred to me to manipulate AccessProtect like that. I feel a little uncomfortable with it—it’s certainly not a black-box test any more, and could theoretically break (or worse, not break when it should) if private implementation details later change. But it does do a very good job of pinning down the problem, and such changes are probably not very likely, so I guess it’s probably fine.

Also a good point about #wakeup needing to be inside a critical section. It turns out that you are absolutely right, and your reproduction still fails, even with my fix, until I additionally wrap #signal in a critical section. (Something about the particular case I encountered in real application code must be slightly less sensitive.) To do so I need a method which is the guts of #cancel, but without a “built-in” critical section—I defaulted to naming it #basicCancel, but I’d be open to better ideas. #unschedule could also be an option—it is basically the opposite of…well, either/both #schedule and #snooze.

Core/Object Arts/Dolphin/Base/Delay.cls
```diff
@@ -44,6 +44,14 @@ Example usage:
 
 	^self resumptionTime <= anotherDelay resumptionTime!
 
+basicCancel
+	^self == Current
```
blairmcg, line 48: The ^-return is presumably from an extract method refactoring, and not actually needed here?

Core/Object Arts/Dolphin/Base/DelayTest.cls
```diff
@@ -0,0 +1,128 @@
...
+DelayTest guid: (GUID fromString: '{0b7c6875-1051-4b80-b137-cd1c02754bcd}')!
+DelayTest comment: ''!
+!DelayTest categoriesForClass!Unclassified! !
+!DelayTest methodsFor!
+
+expectDelay: expectedMilliseconds for: nowMicroseconds
+	| elapsed delta |
+	elapsed := nowMicroseconds - startTime.
+	delta := (elapsed - (expectedMilliseconds * 1000)) abs.
+	self assert: delta < 2000
```
blairmcg, line 17: The variance that can be expected is asymmetric. It should be no more than the resolution shorter (i.e. a 10mS wait won't be less than 9mS), but could be considerably longer.

Core/Object Arts/Dolphin/Base/DelayTest.cls
```diff
@@ -0,0 +1,128 @@
...
+	instanceVariableNames: 'delay startTime'
+	classVariableNames: ''
+	poolDictionaries: ''
+	classInstanceVariableNames: ''!
+DelayTest guid: (GUID fromString: '{0b7c6875-1051-4b80-b137-cd1c02754bcd}')!
+DelayTest comment: ''!
+!DelayTest categoriesForClass!Unclassified! !
+!DelayTest methodsFor!
+
+expectDelay: expectedMilliseconds for: nowMicroseconds
```
blairmcg, line 13: Slightly neater would be to pass a Duration for the expected delay.

Core/Object Arts/Dolphin/Base/DelayTest.cls
```diff
@@ -0,0 +1,128 @@
...
+
+testCancel
+	| proc endTime |
+	delay := Delay forMilliseconds: 20.
+	proc := 
+			[delay wait.
+			endTime := Time microsecondClockValue] forkAt: Processor userInterruptPriority.
+	
+	[delay cancel.
+	self deny: Delay.Current == delay.
```
blairmcg, line 43: `self deny: Delay.Current identicalTo: delay`

blairmcg (approved):
I think you've done a good job. The PR is still in draft mode, but I'm happy to merge it as is. None of my comments need to be actioned.

Regarding the whitebox nature of the test repro of the bug itself, I wouldn't worry about that in this case. Its difficult to construct useful/reliable tests of race conditions without reaching into the internal state. The implementation is stable and relatively unlikely to be changed, but if it were the test is still useful in describing a case, even if it depends on the existing implementation and so would break immediately that implementation were changed.

Core/Object Arts/Dolphin/Base/DelayTest.cls
daniels220: Fair enough—I did it the way I did to avoid repeating "milliseconds" everywhere, but either tradeoff is fine.

Core/Object Arts/Dolphin/Base/DelayTest.cls
daniels220: I don't want to allow it to be *too* considerably longer as that might allow an actual bug to slip through, but I've shifted the window so it's +3ms/-1ms.

daniels220:
Eh, easy enough to make those last couple tweaks. I've done so and marked as not-draft.

PR #1054, opened 2020-11-18
Add binary selector << to String as alias to expandMacros...

### LINKED ISSUE #1053: Binary selector for expandMacrosWith...
When revisiting code I'm gradually replacing string concatenation with `expandMacrosWith:` and its variants. One (very) small annoyance is the need to wrap the modified String expression with parentheses, for example:
```
MessageBox warning: 'Could not locate ', anObject displayString, '. Please check and retry.'
```
...becomes
```
MessageBox warning: ('Could not locate <1d>. Please check and retry.' expandMacrosWith: anObject) 
```

I've implemented a simple workaround for this by adding the binary selector `<<` as an alias for `expandMacrosWithArguments:`. Thus the above example becomes:
```
MessageBox warning: 'Could not locate <1d>. Please check and retry.' << {anObject} 
```

At its most basic this just needs a single method adding to String
```
<< anArray

	^self expandMacrosWithArguments: anArray
```

On checking the base image it can be seen that the majority (>50%) of uses of expandMacros... involve just a single argument. A more sophisticated implementation using double-dispatch could avoid the need to wrap a single argument in an Array whilst still supporting multiple arguments where necessary.

String method:
```
<< anObject

	^anObject expandMacrosIn: self
```
Object method:
```
expandMacrosIn: aString

	^aString expandMacrosWith: self
```
Collection method:
```
expandMacrosIn: aString

	^aString expandMacrosWithArguments: self
```
The Object implementation also needs duplicating in String to avoid a single String argument being treated as a collection of Characters.

This then enables expressions like:
```
'<1d> already exists' << anObject
'<1d> already exists in <2d>' << {anObject. anObject parent}
```
A small downside with this approach is where a collection is intended as an individual argument rather than the collection of arguments. However this can be worked around by wrapping the collection in an Array, e.g.
```
MessageBox warning: 'Collection <1p> too small' << {#(1 2 3)}
```

I'd like to submit this as a potential base enhancement (I'm happy to implement this myself, with associated tests).

rko281:
Add binary selector `<<` to String as alias to `expandMacrosWith:` / `expandMacrosWithArguments:` (double-dispatched from argument). Also add associated tests and fix incorrect reference in comment to `expandMacros` etc. 
Resolves #1053.

### DIFF
--- a/Core/Object Arts/Dolphin/Base/Collection.cls
+++ b/Core/Object Arts/Dolphin/Base/Collection.cls
@@ -323,6 +323,12 @@ errorNotKeyed
 
 	^self error: ('<1p>s do not respond to keyed accessing messages' expandMacrosWith: self class)!
 
+expandMacrosIn: aString
+
+	"Private - Expand aString with the collection of replaceable arguments represented by the receiver."
+
+	^aString expandMacrosWithArguments: self!
+
 fold: aDyadicValuable 
 	"Evaluate the <dyadicValuable> argument, operation, across the elements of the receiver
 	starting with the first two elements. For subsequent evaluations the first argument to the
@@ -629,6 +635,7 @@ union: comperand
 !Collection categoriesFor: #do:separatedBy:!enumerating!public! !
 !Collection categoriesFor: #errorEmptyCollection!exceptions!private! !
 !Collection categoriesFor: #errorNotKeyed!exceptions!private! !
+!Collection categoriesFor: #expandMacrosIn:!double dispatch!private! !
 !Collection categoriesFor: #fold:!enumerating!public! !
 !Collection categoriesFor: #growSize!accessing!private! !
 !Collection categoriesFor: #identityIncludes:!public!searching! !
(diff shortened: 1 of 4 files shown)


Core/Object Arts/Dolphin/Base/Collection.cls
```diff
@@ -323,6 +323,12 @@ errorNotKeyed
 
 	^self error: ('<1p>s do not respond to keyed accessing messages' expandMacrosWith: self class)!
 
+expandMacrosIn: aString
```
blairmcg, line 326: John, this needs to be pushed down to SequenceableCollection, as the String>>expandMacrosWithArguments: method expects an ordered collection with integer keys (it says Array, but any sequence will do). I suppose you could theoretically pass it a Dictionary with integer keys, but that would be a bit perverse. It definitely won't work for Sets, etc.
If you are not in a position to update it today, I can merge and fix later.

rko281: Yes, I won't be able to address this today so if you could merge and fix that would be great - thanks.

PR #1060, opened 2020-11-28
Fix #1059: SmalltalkLocale numberFormat data invalid on image restart

### LINKED ISSUE #1059: SmalltalkLocale numberFormat data invalid on image restart
SmalltalkLocale has a singleton instance that holds a [NUMBERFMTW ](https://docs.microsoft.com/en-us/windows/win32/api/winnls/ns-winnls-numberfmtw) structure that can be used for formatting of Floats into strings. NUMBERFMTW has an unfortunate design where the string values for grouping and decimal separators are not held within the structure as character arrays, but are instead pointers to externally held character arrays. This means that an instance of the structure created in one session is very unlikely to be valid in a subsequent session because the string pointers are either pointing into the ether, or at some random memory. In practice it seems that the former is more usually the case, and access violations occur when the structure is used in a call to [GetNumberFormat](https://docs.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getnumberformatw).

**To Reproduce**

1. Load the regression tests
1. Browser to SmalltalkLocale and run the tests
1. Exit, saving the image
1. Restart
1. Run the tests again
 
**Expected behaviour**
Tests pass both times

**Actual behaviour**
2 test errors running after restart

blairmcg:


Core/DolphinVM/Compiler/compiler.cpp
```diff
@@ -2225,6 +2226,18 @@ void Compiler::ParsePrimitive()
...
+			{
+				NextToken();
+				if (ThisTokenIsBinary('>'))
+				{
+					NextToken();
+					this->m_isMutable = true;
+					return;
+				}
+				else
+					CompileError(CErrExpectCloseTag);
```
blairmcg, line 2239: Needs a test added

PR #1068, opened 2020-12-07
Fix extraneous horizontal scroll bar in ListView with a large number of items (fixes #1066)

### LINKED ISSUE #1066: ListView with very large number of items shows unnecessary horizontal scrollbar
A ListView with at least one auto-sized column and enough items to be >`2**16` pixels high (a little over 1500 items for a `ListView new showShell` with default font etc) acts as though it shouldn't have a vertical scrollbar, thus making the auto-size column too wide and ending up with a *horizontal* scrollbar as well.

This is because `LVM_APPROXIMATEVIEWRECT` returns its answer as a pair of 16-bit integers packed into the 32-bit return value, so at a certain point it overflows and wraps around. I can see two basic approaches:

We could just assume that if we have over, say, 1000 items, the vertical scrollbar will be needed, and don't even bother sending LVM_APPROXIMATEVIEWRECT. Or, as a refinement, calculate the height of a single item (with `itemRect:`), multiply by the number of items, and if this exceeds a conservative value (say `2**15`), again assume a vertical scrollbar.

Or, we could calculate an exact required height by using the difference between the `itemRect:` of the first and last item. For this to work, we also need to account for the column header (it looks like `LVM_APPROXIMATEVIEWRECT` does so internally). The code to do this directly (by querying the unmanaged view that represents the header) exists, but it's in SSW ListView Extensions, not core. But we could also take advantage of the fact that `LVM_APPROXIMATEVIEWRECT` accounts for it: Ask for the height required to display 0 items, then add that to the difference of `itemRect:`s.

Any thoughts on which approach to take?

----

Two minor related details:

`autoResizeColumns:` doesn't specify whether the argument is a client width or full frame width, and the two callers are inconsistent in which they pass. The implementation actually expects a full frame width, but as a comment notes, the adjustment it performs to get a client width is arbitrary and only correct for some combinations of `hasBorder`, `hasStaticEdge`, `hasClientEdge`. This can be replaced by `calcClientRectangleFromRectangle:`, which is correct in all cases AFAICT. I'd tend to make the argument always a proper client width and change `onPositionChanged` to do the calculation.

Well, except that if the view *in fact* has one or both scrollbars, `calcClientRectangleFromRectangle:` does not account for their presence, even though they do affect the answer from `clientRectangle`. For our purposes here (responding to position changes before the view has had a chance to update its scrollbars) this is a good thing, but it seems wrong in general? Or, since this comes straight from the behavior of WM_ADJUSTWINDOWRECTEX, perhaps this is a Win32 gotcha that's nonetheless expected behavior?

**Example **
```Smalltalk
lv := ListView new showShell.
lv viewMode: #report.
lv list: (1 to: 1600).
"Resize the shell just a little bit, observe that both scrollbars appear"
```

daniels220:
Compute the required height for the control, and therefore whether it will have a vertical scrollbar, using a method that will not overflow if the required height exceeds 32768 (2**15).
Also change guard condition to isClientAreaChanged rather than isRectangleChanged—changes to edge styles should cause a re-layout, while simple moves don't need to.

### DIFF
--- a/Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
+++ b/Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
@@ -172,11 +172,12 @@ autoResizeColumns
 
 	self isReportMode ifTrue: [self autoResizeColumns: self clientWidth]!
 
-autoResizeColumns: anInteger
-	"Private - Attempt to share the specified <integer> width between all automagically
-	resizeable columns in the receiver. This only has a visible effect in #report mode."
+autoResizeColumns: clientWidth
+	"Private - Attempt to share the specified <integer> width (which must be the client width of the view,
+	after subtracting for borders and such) between all automagically resizeable columns in the receiver.
+	This only has a visible effect in #report mode."
 
-	| cols autoCols fixedWidth idealWidth remainingWidth clientWidth |
+	| cols autoCols fixedWidth idealWidth remainingWidth |
 	fixedWidth := 0.
 	autoCols := 0.
 	cols := self allColumns.
@@ -186,13 +187,11 @@ autoResizeColumns: anInteger
 				ifTrue: [autoCols := autoCols + 1]
 				ifFalse: [fixedWidth := fixedWidth + (self lvmGetColumnWidth: eachKey - 1)]].
 	autoCols == 0 ifTrue: [^self].
-	"Not sure where the magic number 4 comes from, but it doesn't appear to be correct any more"
-	clientWidth := anInteger - 4.
-	"#1063: ListView doesn't seem to report its actual client width correctly if it has a static edge
-	but not one of the other border styles, hence we must 'adjust' it a little in this case."
-	(self hasStaticEdge and: [(self hasClientEdge or: [self hasBorder]) not])
-		ifTrue: [clientWidth := clientWidth - 2].
 	remainingWidth := clientWidth - fixedWidth.
+	"#1063: If we have a static edge but no other edge style, using the full reported client width
+	results in a scrollbar for some reason. Adjust the available width a little to prevent this."
+	(self hasStaticEdge and: [(self hasClientEdge or: [self hasBorder]) not])
+		ifTrue: [remainingWidth := remainingWidth - 2].
 	remainingWidth <= 0 ifTrue: [^self].
 	"Split the available width between the auto columns - note the use of fractional arithmetic to avoid accumulating a rounding error"
 	idealWidth := remainingWidth / autoCols.
@@ -1759,22 +1758,25 @@ onPositionChanged: aPositionEvent
 	control, as that is when it updates scrollbars which can cause the horizontal scrollbar to
 	briefly flicker if the column widths are not adjusted beforehand (#2103)."
 
-	(aPositionEvent isRectangleChanged and: [self isReportMode])
+	(aPositionEvent isClientAreaChanged and: [self isReportMode])
 		ifTrue: 
-			[| width count |
-			width := aPositionEvent width.
+			[| clientWidth requiredHeight count |
+			clientWidth := (self calcClientRectangleFromRectangle: aPositionEvent rectangle) width.
 			count := self itemCount.
+			"We cannot use LVM_APPROXIMATEVIEWRECT on its own,
+			as it will overflow for large numbers of items. Instead we use it to account for
+			the height of the column header and any borders/edge styles, then compute
+			the height of the items separately."
+			requiredHeight := (self
+						lvmApproximateViewRect: 0
+						cx: aPositionEvent width
+						cy: aPositionEvent height) y
+						- SystemMetrics current scrollbarHeight.
 			count > 0
-				ifTrue: 
-					[| extent height |
-					height := aPositionEvent height.
-					extent := self
-								lvmApproximateViewRect: count
-								cx: width
-								cy: height.
-					extent y - SystemMetrics current scrollbarHeight > height
-						ifTrue: [width := width - SystemMetrics current scrollbarWidth]].
-			self autoResizeColumns: width].
+				ifTrue: [requiredHeight := requiredHeight + (self itemRect: count) bottom - (self itemRect: 1) top].
+			requiredHeight > aPositionEvent height
+				ifTrue: [clientWidth := clientWidth - SystemMetrics current scrollbarWidth].
+			self autoResizeColumns: clientWidth].
 	^super onPositionChanged: aPositionEvent!
 
 onRightButtonPressed: aMouseEvent
--- a/Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
+++ b/Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
@@ -122,6 +122,15 @@ addOrUpdate: aBoolean nonVirtualItems: aSequenceableCollection afterIndex: anInt
 						indent: 0.
 					self lvmSetItem: lvItem]]!
 
+adjustForStaticEdgeBug: aRectangle
+	"#1063: If we have a static edge but no other edge style, the view behaves as though it
+	is slightly smaller than the reported client width for the purposes of calculating if scrollbars
+	are shown. Adjust the available width a little to prevent this."
+
+	(self hasStaticEdge and: [(self hasClientEdge or: [self hasBorder]) not])
+		ifTrue: [aRectangle corner: aRectangle corner - 2].
+	^aRectangle!
+
 allColumns
 	"Answer an <sequencedReadableCollection> of all the columns in the view including
 	the primary column."
@@ -170,7 +179,8 @@ autoResizeColumns
 	"Private - Attempt to share the available width in the receiver between all automagically resizeable
 	columns. This only has a visible effect in #report mode, or if the list is subsequently switched to report mode."
 
-	self isReportMode ifTrue: [self autoResizeColumns: self clientWidth]!
+	self isReportMode
+		ifTrue: [self autoResizeColumns: (self adjustForStaticEdgeBug: self clientRectangle) width]!
 
 autoResizeColumns: clientWidth
 	"Private - Attempt to share the specified <integer> width (which must be the client width of the view,
@@ -188,10 +198,6 @@ autoResizeColumns: clientWidth
 				ifFalse: [fixedWidth := fixedWidth + (self lvmGetColumnWidth: eachKey - 1)]].
 	autoCols == 0 ifTrue: [^self].
 	remainingWidth := clientWidth - fixedWidth.
-	"#1063: If we have a static edge but no other edge style, using the full reported client width
-	results in a scrollbar for some reason. Adjust the available width a little to prevent this."
-	(self hasStaticEdge and: [(self hasClientEdge or: [self hasBorder]) not])
-		ifTrue: [remainingWidth := remainingWidth - 2].
 	remainingWidth <= 0 ifTrue: [^self].
 	"Split the available width between the auto columns - note the use of fractional arithmetic to avoid accumulating a rounding error"
 	idealWidth := remainingWidth / autoCols.
@@ -1760,22 +1766,17 @@ onPositionChanged: aPositionEvent
 
 	(aPositionEvent isClientAreaChanged and: [self isReportMode])
 		ifTrue: 
-			[| clientWidth requiredHeight count |
-			clientWidth := (self calcClientRectangleFromRectangle: aPositionEvent rectangle) width.
+			[| clientRect clientWidth count |
+			clientRect := self
+						adjustForStaticEdgeBug: (self calcClientRectangleFromRectangle: aPositionEvent rectangle).
+			clientWidth := clientRect width.
 			count := self itemCount.
-			"We cannot use LVM_APPROXIMATEVIEWRECT on its own,
-			as it will overflow for large numbers of items. Instead we use it to account for
-			the height of the column header and any borders/edge styles, then compute
-			the height of the items separately."
-			requiredHeight := (self
-						lvmApproximateViewRect: 0
-						cx: aPositionEvent width
-						cy: aPositionEvent height) y
-						- SystemMetrics current scrollbarHeight.
 			count > 0
-				ifTrue: [requiredHeight := requiredHeight + (self itemRect: count) bottom - (self itemRect: 1) top].
-			requiredHeight > aPositionEvent height
-				ifTrue: [clientWidth := clientWidth - SystemMetrics current scrollbarWidth].
+				ifTrue: 
+					[| requiredHeight |
+					requiredHeight := self lvmGetOrigin y + (self itemRect: count) bottom.
+					requiredHeight > clientRect height
+						ifTrue: [clientWidth := clientWidth - SystemMetrics current scrollbarWidth]].
 			self autoResizeColumns: clientWidth].
 	^super onPositionChanged: aPositionEvent!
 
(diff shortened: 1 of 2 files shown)


Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
```diff
@@ -1759,22 +1758,25 @@ onPositionChanged: aPositionEvent
...
-			[| width count |
-			width := aPositionEvent width.
+			[| clientWidth requiredHeight count |
+			clientWidth := (self calcClientRectangleFromRectangle: aPositionEvent rectangle) width.
 			count := self itemCount.
+			"We cannot use LVM_APPROXIMATEVIEWRECT on its own,
+			as it will overflow for large numbers of items. Instead we use it to account for
+			the height of the column header and any borders/edge styles, then compute
+			the height of the items separately."
+			requiredHeight := (self
```
blairmcg, line 1770: Although better than before, it still has the issue that the width is reduced on assumption of a scrollbar before the control actually inserts one. 
A nice way to demonstrate this is to put a `Sound bell` statement inside the if condition that reduces the width on line 1776. If you then work with a test ListPresenter with a couple of items in it and a nice large font to make things more obvious, you'll notice that the clicks start long before the vertical scrollbar actually appears.
I think this may be because the LVM_APPROXIMATEVIEWRECT message is assuming some extra space, maybe for another item. I notice you've reduced the height it returns by the height of a scrollbar, and that just happens to be almost right for a normal sized font, reducing the error to a few pixels, but it gets more noticeable with larger fonts.
Unfortunately the extra height does not seem to be that for a full row either (this can be quickly demonstrated with the `Sound bell` still in place by also removing the expression on line 1774 and changing the `top` at the end of line 1776 to `bottom` ), at which point the bell doesn't start ringing soon enough.

This lead me to think that your alternate option of using the actual header height might be the way to go, but then I'm not sure we even need to know the height of the header, since this is accounted for in the position of the bottom of the items anyway. Assuming that the origin is always zero, then the bottom of the last item tells us the height. After a bit of fiddling around I ended up with this much simpler code:
```smalltalk
onPositionChanged: aPositionEvent
	"We must resize any auto-sizing columns *before* WM_WINDOWPOSCHANGED is passed to the
	control, as that is when it updates scrollbars which can cause the horizontal scrollbar to
	briefly flicker if the column widths are not adjusted beforehand (#2103)."

	(aPositionEvent isClientAreaChanged and: [self isReportMode])
		ifTrue: 
			[| clientRect clientWidth count |
			"Use calcClientRectangleFromRectangle: to account for borders, etc"
			clientRect := (self calcClientRectangleFromRectangle: aPositionEvent rectangle).
			clientWidth := clientRect width.
			count := self itemCount.
			"Adjust for vertical scroll bar if one is going to be needed, which will definitely not be the case if there are no items."
			count > 0
				ifTrue: 
					[| requiredHeight |
					requiredHeight := (self itemRect: count) bottom - self lvmGetOrigin y.
					requiredHeight > clientRect height
						ifTrue: 
							["Sound bell."
							clientWidth := clientWidth - SystemMetrics current scrollbarWidth]].
			self autoResizeColumns: clientWidth].
	^super onPositionChanged: aPositionEvent
```
This seems to be basically correct, but still doesn't handle the case where a horizontal scroll bar reduces the height because (as you pointed out in the issue) this is not accounted for in the client rectangle returned by `calcClientRectangleFromRectangle:`. It shouldn't be too hard to take that into account, although it does mean repeating some of the `autoResizeColumns:` logic..

I am happy to merge your change as is, since the issue referred to is present anyway and could be logged separately.

Just let me know what you want to do.

blairmcg (approved):

Core/Object Arts/Dolphin/MVP/Views/Common Controls/ListView.cls
daniels220: Hah, you're right about the scrollbarHeight thing not working for different font sizes—it was that way in the old code, and I just assumed it was correct without checking more scenarios.

SO! I decided I should seriously check all the scenarios, and...well. Wrote up a workspace to do so—[see gist](https://gist.github.com/daniels220/2999064985d6e8bb9b8625e05d081d25). Sure enough, the scrollbarHeight adjustment basically only works in the default case.

The reason it got as exhaustive as it did is that I found that the view doesn't quite use its `clientRectangle` to decide whether or not scrollbars are shown. Certain edge styles are sliiiightly inconsistent about the cutoff. Some combinations allow the items to extend either 2px or 4px past the nominal bottom of the client area before a scrollbar appears. And interestingly, the one exception to the client *width*—static edge but no border or client edge—applies here as well, with the scrollbar appearing 2px *sooner* than expected.

So...that's a mess. It's possible there's some rational explanation, but if so, I don't know what it is—the only solution I've found is an explicit table of special cases.

----

Re: the header height and such: I confirmed that, in all the cases I checked anyway, the following are true:

```
<LVM_AVR for 1 item> - <LVM_AVR for 0 items> = (self itemRect: 1) height
<LVM_AVR(0)> - <row height> = self headerRect height
thus
<LVM_AVR(0)> - (<LVM_AVR(1)> - <LVM_AVR(0)>) = self headerRect height
or
<LVM_AVR(0)> * 2 - <LVM_AVR(1)> = self headerRect height
```

In other words, we *can* compute the exact header height without using the functions to request it explicitly in SSW ListView Extensions. Looking at it, though, that does seem kinda silly when all the components to do it explicitly are there in core and it's just the helper function that's an extension—might be best to just adopt it.

Using `LVM_GETORIGIN` is clever—I didn't know that existed—although it's actually `(self itemRect: count) bottom + self lvmGetOrigin y`, not minus. The thing is that while this *works*, [the docs explicitly say it shouldn't](https://docs.microsoft.com/en-us/windows/win32/controls/lvm-getorigin): "Returns TRUE if successful, or FALSE if the current view is list *or report view*." So it miiiight not be wise to rely on.

----

So. Let me know what you think about:
1. The value of an exact solution at the cost of horrid explicit special cases. A halfway option would be to handle the static-edge-and-nothing-else case by extracting a common helper between autoResizeColumns: and here, while leaving the cases where we assume vscroll too *soon* as harmless.
2. Adopting `headerRect` (and maybe `headerControl` as well, why not) into core from SSW ListView Extensions, OR
3. Relying on Microsoft not breaking backcompat even though the docs say LVM_GETORIGIN shouldn't work.

And I can piece together a solution (and tests) accordingly.

----

P.S.

> but still doesn't handle the case where a horizontal scroll bar reduces the height

Actually that case is irrelevant here, because...decision tree:

If the fixed columns require more horizontal space than the entire client width, we will always have a horizontal scroll bar—but we also won't do any resizing of the flexible columns, so it doesn't *matter* whether there's a vertical scroll bar or not, we don't need the argument to autoResizeColumns: to be accurate because it won't do anything.

If the fixed columns would/will fit even *with* vscroll, then there will *never* be hscroll (the whole point is to make sure of that) and we just need to decide whether there will be vscroll based on the required height in order to resize appropriately.

If it's right on the edge, so that if we have vscroll we don't have room and will end up with hscroll also, but if we don't then we can resize—so it's all-or-nothing, basically—then we need to decide if the "nothing" case is possible, i.e. if we can do without vscroll *assuming* we don't need hscroll. If it turns out we need vscroll no matter what, we don't really *care* that now we *really* need it because we lose some height to the hscroll as well—as far as autoResizeColumns: is concerned, we're just back in case 1 of not having room for the fixed columns and not doing any work.

Note that the above isn't the in-order logic of the code, but rather a sort of retrospective of the possible outcomes—happily, the code can ignore most of the considerations above and it will all just work out.

Note also that ScrollingDecorator does *not* quite have this luxury, or not in the same way anyway. I've done some work on the logic there that I really need to get together to contribute sometime...

blairmcg: > Hah, you're right about the scrollbarHeight thing not working for different font sizes—it was that way in the old code, and I just assumed it was correct without checking more scenarios.
> 

Yes, it is a pre-existing issue, just one that might go unnoticed. I don't think it has ever registered consciously for me.

I wondered about why it might have been done like that, so did a bit of code archaeology.  git blame points at 7a73688772b31766d10fd29393e21f66cf8344b4, but that is just at a refactoring. The code is basically the same in the origin github source upload, and also in D6. Unfortunately I don't have the ability to see back into the prior history any more.

I suspect it was just a guess. It would have been obvious that some adjustment was needed to the height, and deducting the scrollbar height would have been an obvious thing to try, and apparently worked. 

> SO! I decided I should seriously check all the scenarios, and...well. Wrote up a workspace to do so—[see gist](https://gist.github.com/daniels220/2999064985d6e8bb9b8625e05d081d25). 

I think you added #lvmApproximateViewHeight:, which is needed to run your expressions. I guessed at the implementation.

> Sure enough, the scrollbarHeight adjustment basically only works in the default case.
> 

It seems to be a close approximation, which would tie up with it having been a guess at the adjustment required. It is a related metric, and works closely enough when using the typical default font.

> The reason it got as exhaustive as it did is that I found that the view doesn't quite use its `clientRectangle` to decide whether or not scrollbars are shown. Certain edge styles are sliiiightly inconsistent about the cutoff. Some combinations allow the items to extend either 2px or 4px past the nominal bottom of the client area before a scrollbar appears. And interestingly, the one exception to the client _width_—static edge but no border or client edge—applies here as well, with the scrollbar appearing 2px _sooner_ than expected.
> ...
> In other words, we _can_ compute the exact header height without using the functions to request it explicitly in SSW ListView Extensions. Looking at it, though, that does seem kinda silly when all the components to do it explicitly are there in core and it's just the helper function that's an extension—might be best to just adopt it.

I have no issues with adopting it at all, if @rko281 is happy. In fact I think adopting it is a good idea. However, I don't think it is really needed for this as there is a simpler solution.

> Using `LVM_GETORIGIN` is clever—I didn't know that existed—although it's actually `(self itemRect: count) bottom + self lvmGetOrigin y`, not minus. 

Yes thanks for the correction, it's the scroll offset.

> The thing is that while this _works_, [the docs explicitly say it shouldn't](https://docs.microsoft.com/en-us/windows/win32/controls/lvm-getorigin): "Returns TRUE if successful, or FALSE if the current view is list _or report view_." So it miiiight not be wise to rely on.

I think it will be fine. My reading would be that documentation is more than likely incorrect, and reflects a misinterpretation of either the spec or the original code by the technical writer. There is no obvious reason why the scroll origin would be hard to calculate in any of the modes, or why the developer would have made some of the modes special cases - if anything it seems like that would require extra code. Even if it was deliberate for some unobvious reason, Microsoft are very unlikely to change it now since the old common controls are in maintenance mode. 

On the other hand it is very likely that at some point the geometries will change as part of a refresh of the Windows theme. Actually it is quite likely that the geometries are already different in alternate themes, although I haven't checked. Therefore it would be best to avoid doing anything that requires using magic constants deduced from your extensive examination, as interesting as that is 😊. Sometimes one has to do that as there is no alternative, but it looks like we can keep it simple here. 

> 
> So. Let me know what you think about:
> 
> 1. The value of an exact solution at the cost of horrid explicit special cases. A halfway option would be to handle the static-edge-and-nothing-else case by extracting a common helper between autoResizeColumns: and here, while leaving the cases where we assume vscroll too _soon_ as harmless.

I would err on the side of ignoring the few corner cases you highlight where the calculation assumes a vertical scrollbar will be needed, but the control doesn't actually add one, even though final row is not completely visible. If I understand them correctly, these cases are unlikely to be encountered, being a combination of edge styles and specific heights close to the borderline between needing or not needing a vertical scroll bar. Also when these cases are encountered the impact is very minimal, i.e. just not using the full available width by the small margin of the width of a scrollbar. That doesn't justify encoding a complex set of adjustments based on empirical observations that have no logical pattern, especially as these cases are probably bugs in the controls' own internal sizing logic. 

In summary, the `likelihood of encounter * impact when encountered` measure makes it a won't fix, even before considering the cost and high chance it will get broken by future changes.

> 2. Adopting `headerRect` (and maybe `headerControl` as well, why not) into core from SSW ListView Extensions, OR
> 3. Relying on Microsoft not breaking backcompat even though the docs say LVM_GETORIGIN shouldn't work.
> 

As I said I have no issue with adopting the header extensions, but the LVM_GETORIGIN solution is neat and simple and avoids the need to consider the header, so I'd go for a minimal implementation using (3) every day, even if there are still some rare corner cases.

> And I can piece together a solution (and tests) accordingly.

It would be a good idea to add a simple regression test (perhaps not individually as the view tests have high setup overhead, but as an appropriate assertion at the top of one of the others) specifically to check for Microsoft changing LVM_GETORIGIN to fail per the documentation when in report mode.

> 
> P.S.
> 
> > but still doesn't handle the case where a horizontal scroll bar reduces the height
> 
> Actually that case is irrelevant here, because...decision tree:
> 

Yes, true. If I'd stopped to think a bit more I'd have realised that the point of resizing the columns is to fit within the available width, and therefore there should only be a horizontal scroll bar if there is insufficient width for all the fixed width columns, in which case we don't really care anyway.

Thanks

Blair

daniels220: > Microsoft are very unlikely to change it now since the old common controls are in maintenance mode

Yeah, that's the bottom line, isn't it. Seems reasonable.

> ... Actually it is quite likely that the geometries are already different in alternate themes ... I would err on the side of ignoring the few corner cases you highlight

Right, good point about themes, and yeah, overall not worth the bother. I got a bit carried away investigating, I admit—I was *hoping* to find some internal logic to it that I could make use of, but no such luck. I've adopted more-or-less your implementation from your previous comment.

The one thing that *does* seem worth accounting for is the 2px *reduction* in effective clientExtent when we have static edge but no other edge style. That has remained accurate since D6, and the symptom is slightly higher-impact (we end up showing a horizontal scroll bar when we shouldn't). So, I've moved that correction into a helper method that deals with both width and height.

> It would be a good idea to add a simple regression test

Now that I think about it, I don't think a separate regression test, or even assertion, is necessary at all, since the symptom of Microsoft changing LVM_GETORIGIN would be a `Common control call failed` error every time `onPositionChanged:` is called.

blairmcg: > Now that I think about it, I don't think a separate regression test, or even assertion, is necessary at all, since the symptom of Microsoft changing LVM_GETORIGIN would be a `Common control call failed` error every time `onPositionChanged:` is called.

Yes, that's true.

blairmcg: @daniels220, unfortunately I think the LVM_GETORIGIN call does have a problem. A good way to see it is to open a package browser and select the root node so that all the packages are listed in the package list below the tree. Then scroll down the list to move the list scroll origin. Now select some node in the tree that has 1 or a few packages under it. The list will appear to be empty, but isn't. Moving the mouse cursor down to the bottom of the list should eventually cause some rows to be displayed, but offset.

image

Removing the LVM_GETORIGIN call resolves the issue. It seems to mess up the internal state of the ListView and the vertical scroll origin is not cleared when the contents of the list are refreshed. This might be something to do with timing, or it might be why the docs state it doesn't work in list/report mode. Either way I think we will need to use the `headerRect` after all. 

Sorry about that.

blairmcg: Actually, it's not the LVM_GET ORIGIN call that causes the problem. Or anything else in the calculation. 

I'd reverted the whole method, and it is in fact the change to respond only to client area changes, and not rectangle changes.

I spent a little time looking into why, and the problem appears related to the fact that ListView>>basicRefreshContents sends an LVM_SETITEMCOUNT message to prep the view to receive a particular number of items. This results in receipt of a position changing message with the SWP_FRAMECHANGED flag set. I'm not sure why, although it might be a way of sending a notification that scrollbars may be added or removed. If this flag is used to trigger a column resizing, then the it seems the control's state internal scroll state becomes corrupted for some reason. 

If I change the test from `isClientAreaChanged` to `isResize` (or the original `isRectangleChanged`), then the problem goes away, but your tests of the column width responding to frame style changes start to fail as the column resizing is no longer responding to frame changes.

PR #1075, opened 2020-12-14
Fix corrupted ListView scroll state when reducing item count (fixes #1071)

### LINKED ISSUE #1071: ListViews failing to paint correctly when refreshed for a smaller set of items
PR##1068 has caused a regression in ListView behaviour which can be seen when moving from a long to a shorter list, e.g. in class or package browsers. This is not to do with the changes in the calculation, but rather due to the calculation now being performed when #onPositionChanged: is sent for a SWP_FRAMECHANGED notification. The control sends such a position changing notification (i.e. a WM_WINDOWPOSCHANGING message) when it is sent an LVM_SETITMECOUNT message, as this is how the number of items to display in a ListView in virtual mode is communicated to the control. It seems that responding to the position changing event in this case with operations that affect the view geometry causes the control's internal state to become corrupted. Specifically the scroll origin gets set to an incorrect value such that the displayed rows are offset outside the painted area.

The limitations as to what is allowed in a WM_WINDOWPOSCHANGING handler are not documented for SysListView, but it is not that unusual for such limitations to be present.

If the ListView>>onPositionChanging: handler is modified so that the auto-column sizing is only performed on actual resizes, then the bug does not occur. However, this has other consequences (such as not responding to real frame size changes, e.g. from changing edge styles), which causes some of the new tests to fail. 

Historically the ListView integration did not correctly adjust column widths dynamically when edge styles were changed, and it would not be unreasonable to revert to that, although obviously the tests will need some adaptation.

**Repro Steps**

1. Open a class browser
2. Select Object
3. Scroll down the method list to the bottom
4. Select Animal - note that no methods are visible
5. Select Object again - no methods are visible, until one scrolls down the list

At step 5 it is instructive to use the visual object finder to open an inspector on the method list view, and then evaluate a `self lvmGetOrigin` to see the scroll origin, which will have a fairly large negative 'y' component when it should be zero.

image

daniels220:
This gets the two skipped tests working again, and handles responding to added, resized or removed columns. I decided against overriding the edge-style setters to add an explicit `self autoResizeColumns` as this seems like much more of an edge case, though that is an option as well.

### DIFF
--- a/Core/Object Arts/Dolphin/MVP/ListViewTest.cls
+++ b/Core/Object Arts/Dolphin/MVP/ListViewTest.cls
@@ -111,7 +111,10 @@ testAutoResizeColumns
 	self verifyColumnWidths: #(48 100 48).
 	"Rounding errors do not accumulate"
 	presenter columnsList second width: 101.
-	self verifyColumnWidths: #(48 101 47)!
+	self verifyColumnWidths: #(48 101 47).
+	"Resizes when a column is removed"
+	presenter removeColumnAtIndex: 3.
+	self verifyColumnWidths: #(95 101)!
 
 testAutoResizeColumnsPositionChangedAddingScrollbar
 	| shellNCHeight |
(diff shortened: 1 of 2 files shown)


blairmcg:
> I decided against overriding the edge-style setters to add an explicit `self autoResizeColumns` as this seems like much more of an edge case, though that is an option as well.

I was thinking along the same lines.

PR #1087, opened 2021-01-27
Speed up rebuild of SmalltalkSystem message selector cache

blairmcg:
SmalltalkSystem maintains a cache of all implemented messages in the system that is used (e.g.) for auto-completion. The cache is reset when either a new method is added, or removed, or a class is removed. Rebuilding the cache is generally fast (sub-50ms), but might take long enough to cause a noticeable delay in a very large image with a lot of classes and methods. We can easily speed up the the underlying Behaviour>>understoodSelectorsDo: method and about halve the time the cache rebuild takes.


blairmcg:
Needs a bit of rework

PR #1100, opened 2021-02-17
Replace deprecated function call

jgfoster:
Use `KernelLibrary` instead of `VMLibrary` for `isWindowsVistaOrGreater`.


blairmcg:
For some reason github didn't receive the branch build success notification from appveyor for this, but I can see it worked so I'll merge it. I'll also cherry-pick this down into release/7.1. If working in 7.1 you will want to target that for PR's at present as they generally won't merge into master (this case is lucky - any addition of classes/methods probably won't be). I'll be merging back changes from release/7.1 to master regularly.

PR #1146, opened 2021-07-22
Bv consistency004

brickviking:
I'm interested in your thoughts on the first commit c80e8a539b52339f0019a3b66519b9bf0524afdb, relating to UI.ListView lvmRedrawItems:to:, as the comment originally appeared to be copied verbatim from lvmInsertItem: but doesn't appear to match what lvmRedrawItems:to:  actually does. I know for experienced Smalltalkers, the code's probably pretty obvious but it's not quite so for me, hence the attempt at a comment.

Most of the rest of these are simple typos, though there's a lot of examples of "xxx receivers ..." when it probably should be "xxx receiver's ..." instead. I haven't collected all of those yet, as there's over two hundred when I last checked.


Core/Object Arts/Dolphin/MVP/Base/OS.SCROLLINFO.cls
```diff
@@ -35,7 +35,7 @@ fMask: anInteger
 	bytes dwordAtOffset: _OffsetOf_fMask put: anInteger!
 
 maskIn: anInteger
-	"OR in to contents' mask member the bit flags represented by anInteger.
+	"OR in to content's mask member the bit flags represented by anInteger.
```
blairmcg, line 38: OR in to the mask field the bit flags ...

Core/Object Arts/Dolphin/MVP/Presenters/List/UI.ListBox.cls
```diff
@@ -400,7 +400,7 @@ maybeSelChanging: aSymbol
 	self onSelChanged: curSel!
 
 newMultiSelectionsFromEvent: aMouseEvent
-	"Private - Answer a collection of the <integer> selections that would occur if the `MouseEvent` argument was was passed to the list box control.
+	"Private - Answer a collection of the <integer> selections that would occur if the `MouseEvent` argument was passed to the list box control.
```
blairmcg, line 403: This method comment is redundant and should just be deleted.

Core/Object Arts/Dolphin/MVP/Views/Common Controls/OS.CCITEM.cls
```diff
@@ -48,7 +48,7 @@ iImage: anObject
 	self subclassResponsibility!
 
 image: anInteger
-	"Set contents' iImage member to anInteger"
+	"Set content's iImage member to anInteger"
```
blairmcg, line 51: Set the iImage field to ...

Core/Object Arts/Dolphin/MVP/Views/Common Controls/OS.CCITEM.cls
```diff
@@ -87,7 +87,7 @@ mask: anInteger
 	bytes dwordAtOffset: _OffsetOf_mask put: anInteger!
 
 maskIn: anInteger
-	"OR in to contents' mask member the bit flags represented by anInteger"
+	"OR in to content's mask member the bit flags represented by anInteger"
```
blairmcg, line 90: Same treatment for this and all the other cases below of "content's blah member", translate to "the blah field"

Core/Object Arts/Dolphin/MVP/Views/Common Controls/OS.CCITEM.cls
```diff
@@ -165,7 +165,7 @@ text
 	^text ifNil: [self pszText]!
 
 text: aString
-	"Sets contents' pszMember to Utf16String. We also hang on to it in our text instance var to
+	"Sets content's pszMember to Utf16String. We also hang on to it in our text instance var to
```
blairmcg, line 168: Set the pszMember field to...

Core/Object Arts/Dolphin/MVP/Views/Common Controls/OS.LVCOLUMNW.cls
```diff
@@ -173,7 +173,7 @@ text
 	^text!
 
 text: aString
-	"Sets contents' pszMember to a Utf16String. We also hang on to it in our
+	"Sets content's pszMember to a Utf16String. We also hang on to it in our
```
blairmcg, line 176: Set the pszMember field to ...

Core/Object Arts/Dolphin/MVP/Views/Common Controls/UI.ListView.cls
```diff
@@ -1318,7 +1318,7 @@ lvmInsertItem: aLvItem
 		lpParam: aLvItem asParameter!
 
 lvmRedrawItems: firstIndex to: lastIndex
-	"Private - Insert aLvItem into the tree view and add it to our contents dictionary"
+	"Private - Cause redraw messages to be sent for each of these item entries"
```
blairmcg, line 1321: This is about right, but there is only really any value in comments in these methods that are just sending the windows message if it adds something over and above the Microsoft documentation. In this case it doesn't, so I would just delete it.

Core/Object Arts/Dolphin/MVP/Views/Common Controls/UI.ListView.cls
```diff
@@ -1894,7 +1894,7 @@ primaryColumn
 	^self allColumns first!
 
 queryCommand: aCommandQuery
-	"Update aCommandQuery to indicates how a command would be processed.
+	"Update aCommandQuery to indicate how a command would be processed.
 	if sent to the receiver. Answers whether the receiver recognised the command
 	as one of its own (this may be ignored by the command router). This need not
```
blairmcg, line 1899: The last sentence here, beginning "This need not be sent..", is confusing in this context as it is guidance from the superclass implementation in View as to how overrides should be implemented. It should not have been retained with the rest of the copied down content.

Core/Object Arts/Dolphin/MVP/Views/Common Controls/UI.ListView.cls
```diff
@@ -2628,7 +2628,7 @@ onStartup
 
 	NtDll isWine
 		ifTrue: 
-			["Having autoSelectPackages option turned on in System  Browser gives unpleasant scrolling effect when many multiple packages are selected. This also happens in Package Browser. Problem under Wine traced to LVIS_FOCUSED option in ListView>>selectIndex:. Not sure what overall effect it will have turning this off."
+			["Having autoSelectPackages option turned on in System Browser gives unpleasant scrolling effect when many multiple packages are selected. This also happens in Package Browser. Problem under Wine traced to LVIS_FOCUSED option in ListView>>selectIndex:. Not sure what overall effect it will have turning this off."
```
blairmcg, line 2631: When running under Wine the autoSelectPackages option causes an unpleasant scrolling effect when many packages are selected in a System Browser or Package Browser. The problem does not occur without the LVIS_FOCUSED style, the absence of which does mean the loss of focus indication.

Core/Object Arts/Dolphin/MVP/Views/Common Controls/UI.TabView.cls
```diff
@@ -89,7 +89,7 @@ calcClientRectangleFromRectangle: aRectangle
 calcRectangleFromClientRectangle: aRectangle 
 	"Private - Given a client rectangle represented by the <Rectangle> argument, answer the
 	window rectangle that would be required to achieve this taking into account the current
-	window styles etc Note that the answered rectangle is in the COORDINATE SYSTEM OF THE
+	window styles etc. Note that the answered rectangle is in the COORDINATE SYSTEM OF THE
```
blairmcg, line 92: Not sure why this is shouting. Perhaps somebody got burned by that at the time, but I think we should lose the capitals.

Core/Object Arts/Dolphin/MVP/Views/SysLink/OS.LITEM.cls
```diff
@@ -94,7 +94,7 @@ mask: anInteger
 	bytes dwordAtOffset: _OffsetOf_mask put: anInteger!
 
 maskIn: anInteger
-	"OR in to contents' mask member the bit flags represented by anInteger"
+	"OR in to content's mask member the bit flags represented by anInteger"
```
blairmcg, line 97: Another one of those 'content's mask member' -> 'the mask field' cases

Core/Object Arts/Dolphin/MVP/Base/OS.SCROLLINFO.cls
brickviking: How about ...the receiver's mask field the bit flags... ?

Core/Object Arts/Dolphin/MVP/Views/Common Controls/UI.ListView.cls
brickviking: So, remove the last sentence, or space it out from the previous comment?

Core/Object Arts/Dolphin/MVP/Views/Common Controls/UI.ListView.cls
brickviking: Much better. I'll stick that into place as is.

Core/Object Arts/Dolphin/MVP/Views/Common Controls/UI.ListView.cls
brickviking: Right, I knew it was either this comment or deletion. Deletion it'll be.

Core/Object Arts/Dolphin/MVP/Presenters/List/UI.ListBox.cls
brickviking: It would be good to have a top-level "This is what the code does" not just "this is what each section does".

blairmcg: In the specific case of this method, it is has been factored out of another method that has the same comment and has a name that, when considered in context, describes what it is intended to do better than the comment.

More generally, the code should be clear enough that a comment is not needed to explain what it does. Comments can be a code smell that the code is not clear by itself; this is particularly true of sectional comments, which suggest a block of code should be factored out to a separate method with a name that replaces the comment. Comments also have a tendency (since they are technically redundant information that is not executed and testable) to be wrong and not to keep pace with change in the code. 

The general principles now intended to be applied are:
1) Public APIs must have a comment describing the what something does (not how).
2) Private APIs should generally not have a comment. If they need one, the method name is probably not clear enough.**
3) In some cases some explanation of non-obvious implementation details may be needed, but only if the intent is not easy to express in the code and would otherwise be lost.
4) Any description of the "how" should form part of a separate comment to the initial "what", but always bearing in mind (3).

** For auto-generated code different rules would apply since the maintenance issues are different, and it isn't generally possible to know at generation time whether there is value in the documentation accompanying the artefact from which the code is being generated.

Different principles have been applied over the years. The code has been added by a variety of authors with different views, and those views have changed over time. I really don't think there is sufficient value in changing all that in isolation, rather it makes sense to clean up when working in the area. In fact I would prefer not to change it unless the code is being worked on, as the style of commenting is sometimes useful historical context. Correcting all the typos, grammatical errors, and obviously wrong/irrelevant ones is fine of course.

brickviking:
Right, that should hopefully be the changes I need to make. Hope you don't mind my slight rewrites.

Core/Object Arts/Dolphin/MVP/Base/OS.SCROLLINFO.cls
blairmcg: I'm not especially bothered whether it is "OR into the mask field" or "OR into the receiver's mask field"

Core/Object Arts/Dolphin/MVP/Views/Common Controls/UI.ListView.cls
blairmcg: Remove. The comment is only relevant in the View definition.

PR #1172, opened 2022-12-30
Fix break caused by loss of stdext::hash_value in VS 17.4

jgfoster:
I can't figure out what changed, but I'm no longer able to build Dolphin using VisualStudio 17.4.3 on Windows 10 or Windows 11. I get an error that hash_value is not found in stdext. I've spent a bit of time learning about this function and looking at how Dolphin uses it, and come up with a one-line change that builds and seems to pass the tests. Comments are welcome especially since if this isn't the right fix then I need another fix!

### DIFF
--- a/Core/DolphinVM/objmem.cpp
+++ b/Core/DolphinVM/objmem.cpp
@@ -271,9 +271,21 @@ bool __stdcall ObjectMemory::inheritsFrom(const BehaviorOTE* behaviorPointer, co
 
 #pragma code_seg(GC_SEG)
 
+namespace stdext {
+	template <class _Kty>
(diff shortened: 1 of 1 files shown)


blairmcg:
What happened is that [Microsoft deprecated it in VS17.3](https://github.com/microsoft/STL/pull/2996), so you must have done a VS upgrade James. That's how encountered it a couple of weeks ago anyway.
I fixed this in the master branch with a1d94acee0eae38512f8839569d87e41448a8e84. It would be better to cherry-pick in that specific commit because it is guaranteed to preserve the existing behaviour (I made a copy of the short deprecated function), and it will keep the history common. I just tried it, and it auto-merges correctly.

